PR #45, opened 2013-02-16
C++ support for EUMM

vovkasm:
This patch adds C++ support for EUMM. It introduse parameter (and Makefile macro) with name `XSTARGET_EXT` - it is suffix for files prodused from .xs (.c by default).

This allows modules written in C++ does not use various hacks in order to run the C++ compiler for ".с" files.

I personally test this change with:
1. linux
2. strawberry perl 64 on Win7

Only warning. I blindly made changes in `ExtUtils::MM_VMS` (because I personally can't test for this platform)


vovkasm:
As #195 provides better solution, this pull request can be freely closed.

PR #135, opened 2014-08-11
Handle BEGIN{} around $VERSION in parse_version

mohawk2:
Test added by @jhthorsen. Alternative solution, replacing get_version, by @shadowcat-mst.


karenetheridge:
Module::Metadata now handles this properly, as of 1.000029-TRIAL.

PR #179, opened 2014-12-21
XS 2/3: XS refactoring

mohawk2:
Incorporates therefore close #75.

### DIFF
--- a/lib/ExtUtils/MM_AIX.pm
+++ b/lib/ExtUtils/MM_AIX.pm
@@ -6,9 +6,6 @@ our $VERSION = '7.05_05';
 require ExtUtils::MM_Unix;
 our @ISA = qw(ExtUtils::MM_Unix);
 
-use ExtUtils::MakeMaker qw(neatvalue);
-
-
 =head1 NAME
 
 ExtUtils::MM_AIX - AIX specific subclass of ExtUtils::MM_Unix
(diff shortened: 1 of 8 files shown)


nrensen:
Picking up a comment from https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/192...

I noticed the inconsistency of quoting %1s.bs in some places but not others in MM_Unix.pm:
```
%1$s.bs : $(FIRST_MAKEFILE) $(BOOTDEP)
        $(NOECHO) $(ECHO) "Running Mkbootstrap for %1$s ($(BSLOADLIBS))"
        $(NOECHO) $(PERLRUN) \
                "-MExtUtils::Mkbootstrap" \
                -e "Mkbootstrap('%1$s','$(BSLOADLIBS)');"
        $(NOECHO) $(TOUCH) "%1$s.bs"
        $(CHMOD) $(PERM_RW) "%1$s.bs"

%2$s : %1$s.bs %3$s
        $(NOECHO) $(RM_RF) %2$s
        - $(CP_NONEMPTY) %1$s.bs %2$s $(PERM_RW)
```
I consider it would be better to remove the quoting altogether rather than a token effort that ultimately doesn't work reliably. At the very least I think it should be consistent.

As I understand GNU make does support spaces in targets, but requires more elaborate quoting. I'm not sure whether BSD make supports spaces in targets, I couldn't find a way, but if it does it requires something different from gmake:

```
$ cat Makefile
test1: "foo\ bar"

"foo\ bar":
        @echo target: $@

test2: "foo bar"

"foo bar":
        @echo target: $@

test3: foo bar

foo bar:
        @echo target: $@

$ gmake test1
target: foo bar

$ gmake test2
/bin/sh: no closing quote
Makefile:9: recipe for target '"foo' failed
gmake: *** ["foo] Error 1

$ gmake test3
target: foo
target: bar

$ make test1
/bin/sh: no closing quote
*** Error 1 in /tmp (Makefile:4 '"foo\': @echo target: "foo\)

$ make test2
/bin/sh: no closing quote
*** Error 1 in /tmp (Makefile:9 '"foo': @echo target: "foo)

$ make test3
target: foo
target: bar
```

Quite probably I'm due for some education in the art of make and apologise in advance if this is just noise.

mohawk2:
BSD make doesn't seem to support spaces in dependencies. If you can point me at the canonical source for its source code, I will become better-informed.

GNU make does, if you \ the spaces. dmake does if you surround the item with "", BUT it doesn't do string-concatenation (until I get round to finishing and releasing dmake 4.13, which will).

Within-distro items aren't very assiduously space-quoted because naming them with spaces isn't a meaningful thing, so the fairly extensive tests for space-including directories won't have caught them. That doesn't mean that quoting them is unnecessary. The space-protection is necessary when referring to items outside the distro, such as Perl itself which might live in "Program Files", or the directory in which the distro is unpacked, such as a Windows user's "C:\Documents and Settings" subdir where their temp directory lives.

nrensen:
The source code for OpenBSD's make is here: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/make/

NetBSD, FreeBSD, ... each have their own, so nominating a canonical source is probably contentious.

However, I'm concerned that this is becoming an unhelpful rabbit hole. As it stands the quoting won't have any adverse impact on use cases I care about. If it is necessary for other cases then that's the end of it. Sorry for the noise.

mohawk2:
@nrensen thanks for the extra eyes and the good insight you provided. Not noise at all!

PR #180, opened 2014-12-22
Eliminate && from Makefile recipes - totally unnecessary.

mohawk2:
Justification: increases recipes' portability, notably on Win95.

Split out from #176, since @bulk88 disagrees: https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/168#issuecomment-67590474 and https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/176#issuecomment-67891163.

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -3869,7 +3869,8 @@ sub xs_c {
     return '' unless $self->needs_linking();
     '
 .xs.c:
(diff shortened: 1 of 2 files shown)


mohawk2:
On the point of performance, I ran EUMM's tests (thinking of this as a decent proxy for running builds of EUMM projects) through on Windows once on the code before this change is applied, and twice with it.

Time taken before the change: 334 secs
Time taken after the change: 339, 312 secs

I think you'd get more benefit from using GNU make's `.ONESHELL` special target.

PR #187, opened 2014-12-27
Meta 2/2: stash metadata files in files, not Makefile

mohawk2:
Branched from therefore close #186. Incorporates and therefore: close #90, close #93.


Leont:
I don't think the second set of these patches got the peer review it should have gotten

PR #199, opened 2015-01-20
Prefer stringify() to normal()

JohnPeacock:
More of the same.

### DIFF
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -1424,7 +1424,7 @@ sub _normalize_version {
   $version = 0 unless defined $version;
 
   if ( ref $version eq 'version' ) { # version objects
-    $version = $version->is_qv ? $version->normal : $version->stringify;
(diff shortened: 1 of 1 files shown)


Leont:
@dagolden pointed out this is problematic as it can result in a non-valid version in the meta-data. Quite possibly we should revert this.

mohawk2:
I applied it after you said you thought it was a good idea. If you're changing your mind, feel free to revert it, but I'd suggest a fuller discussion on #toolchain.

karenetheridge:
Whatever happens, I think we need more tests of normal and weird versions
to be sure of the behaviour going forward.

JohnPeacock:
On 01/28/2015 12:21 AM, Karen Etheridge wrote:
> Whatever happens, I think we need more tests of normal and weird versions
> to be sure of the behaviour going forward.

I thought it was a good idea too, obviously, but David is right.  I need 
to fix the alpha->normal behavior for real and haven't had a second to 
look at it since this all went down.

John

PR #206, opened 2015-02-14
do proper conversion to 2.x for META_ADD/MERGE

haarg:
Use CPAN::Meta::Converter to convert to 2.x instead of failing to
reimplement it.  Specifically, this fixes converting resources.
Requiring CPAN::Meta for this bit isn't a problem because everything
using the metafile_data method already requires it.

Passing everything through CPAN::Meta::Converter has the side effect of
cloning all of the structures.  This means we no longer modify the items
in META_ADD when running metafile_data, so running in multiple times
won't cause problems.


karenetheridge:
I've tested the patch against libwww-perl@606a520112f (where I noticed the problem originally) and it works.

PR #211, opened 2015-03-18
for dmake use makefile.mk instead of Makefile

bulk88:
This is an idea from my recent work at optimizing dmake. It also allows 2 makefiles, one for VC (almost always nmake, very rarely dmake) Perl, and one for Strawberry (always dmake) perl to coexist in the same directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *~
 *.old
 Makefile
+makefile.mk
 blib/
 pm_to_blib
 Build
--- a/lib/ExtUtils/MM_Win32.pm
+++ b/lib/ExtUtils/MM_Win32.pm
@@ -141,6 +141,9 @@ sub init_tools {
 
     $self->{NOOP}     ||= 'rem';
     $self->{DEV_NULL} ||= '> NUL';
+    $self->{FIRST_MAKEFILE} ||=
+	$self->{MAKEFILE}
+	|| ($self->is_make_type('dmake') ? 'makefile.mk' : 'Makefile');
 
     $self->{FIXIN}    ||= $self->{PERL_CORE} ?
       "\$(PERLRUN) $self->{PERL_SRC}\\win32\\bin\\pl2bat.pl" :
--- a/t/lib/MakeMaker/Test/Utils.pm
+++ b/t/lib/MakeMaker/Test/Utils.pm
@@ -195,7 +195,7 @@ should generate.
 =cut
 
 sub makefile_name {
-    return $Is_VMS ? 'Descrip.MMS' : 'Makefile';
+    return $Is_VMS ? 'Descrip.MMS' : (make() =~ /\bdmake\b/) ? 'makefile.mk' : 'Makefile';
 }
 
 =item B<makefile_backup>


mohawk2:
EUMM is supposed to make a makefile, not to work alongside other makefiles. I would say this is a bad idea.

somian:
Hauddhi.  Guud Dae.

On Wed, Mar 18, 2015 at 5:43 AM, bulk88 <notifications@github.com> wrote:
> This is an idea from my recent work at optimizing dmake. It also allows 2
> makefiles, one for VC (almost always nmake, very rarely dmake) Perl, and one
> for Strawberry (always dmake) perl to coexist in the same directory.
>

Attention to dmake builds is an interest of mine. By "attention" I
mean of course "that EUMM be optimally aware of dmake".

I suppose that this patchset might include making the generated
makefile a file named "makefile.mk"?  If not, then I have to echo:

> EUMM is supposed to make a makefile, not to work alongside other makefiles. I would say this is a bad idea.

Not only a bad idea, but completely out of place for EUMM.

However, from looking briefly at the patch, I cannot say that the
changes don't simply generate a makefile.mk (there is definitely a
provision in MakeMaker for chosing a different name for the output
makefile).


> Commit Summary
>
> for dmake use makefile.mk instead of Makefile
>
> File Changes
>
> M .gitignore (1)
> M lib/ExtUtils/MM_Win32.pm (3)
> M t/lib/MakeMaker/Test/Utils.pm (2)
>
> Patch Links:
>
> https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/211.patch
> https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/211.diff
>


-- 
Are you concerned that our children are not taught right from wrong?
Why not start by setting an example: HONOR INDIAN TREATIES.

bulk88:
Mohawk, both makefiles are created by EUMM. The "other makefile" is also a EUMM makefile, but for another make tool. EUMM makefiles are always specific to the particular maketool (usually from %Config) and this will never change, right? This patch could also be extended in the future to use generate "GNUmakefile" for gmake platforms.

mohawk2:
What are you trying to achieve by having two makefiles around? Why not keep them elsewhere and copy them in as appropriate?

somian:
On Wed, Mar 18, 2015 at 5:34 PM, bulk88 <notifications@github.com> wrote:

> th
> ​e​
> makefiles are created by EUMM. The "other makefile" is also a EUMM
> makefile, but for another make tool. EUMM makefiles are always specific to
> the particular maketool (usually from %Config) and this will never change,
> right? This patch could also be extended in the future to use generate
> "GNUmakefile" for gmake platforms.


​bulk88: I understood you, as moha​wk did not (uh, clearly, being
objective).  So these makefiles are not crearted by direct editing and are
not permanently around. So they in fact are completely within the MakeMaker
framework. I like it!  Still, it is a little hard to see what you want to
achieve by it.  The alternate `make` tools all understand the list of
primary makefiles includes the one named "Makefile".  And when you use
another `make` it usually means you are using another platform, most
likely. In any event, it is assumed that all users will `make clean; perl
./Makefile.PL` each time they build the pkg. So, what difference does it
make whether these different Makefiles are generated?  You are assumed to
be building for only one platform at a time (to restate the point i just
made).

This is not an idea for users? [am guessing]. It's an idea for developers,
tho.  To help them debug MakeMaker's generation by getting different
Makefiles created, comparing them, maybe?

I am really on your side bulk88 because i have done a lot of MakeMaker
tweaking in the past.  Maybe a lot more (on Cygwin, mswin32, linux) than
others.  But you need to make a checklist to document why you think this
idea would be useful to anyone but yourself.  I'd start by thinking about
your workflow. then being very concise, and think about making a project
page where this idea is presented with easily browseable code. For example
it takes only a few minutes to create a github project.

   Regards!

-- 
*https://plus.google.com/+SörenAndersenTinoTheTwalker/
<https://plus.google.com/+S%C3%B6renAndersenTinoTheTwalker/>*
somian08@gmail.com

Are you concerned that our children are not taught right from wrong?  Why
not start by setting an example: *HONOR INDIAN TREATIES*.

haarg:
This would break CPAN.pm.  It expects Makefile.PL to produce a Makefile.

dolmen:
If the aim is to block `nmake` from running a Makefile generated for `dmake`, there should be a better way to do it. What about injecting code in the Makefile that would check the `make` running?

haarg:
The intention is not to block the Makefile from running, but to allow both versions to exist simultaneously.  This is an extremely limited use case, and will break a number of things.

A user can already specify the name of the Makefile to use if they desire. `perl Makefile.PL MAKEFILE=makefile.mk`

dolmen:
And even: `set PERL_MM_OPT=MAKEFILE=makefile.mk`

somian:
​About makefile.mk for​ dmake ...:

On Tue, Jun 9, 2015 at 8:08 AM, Graham Knop <notifications@github.com>
wrote:

> This is an extremely limited use case


​That is clearly the case.  I was hoping that the OP would provide a bit of
example to show why dmake code that was noncompatible with what is
ordinarily generated would be useful to him.  That didn't happen but the
thread is still attracting comments (which i am glad of.  sometimes
suggestions are rejected prematurely ... so just on principle ...although
this one IS a bit edgy)

    Ciao​


-- 
*https://plus.google.com/+SörenAndersenTinoTheTwalker/
<https://plus.google.com/+S%C3%B6renAndersenTinoTheTwalker/>*
‟I like my lovers supple, my coffee strong, and my daimoku outdoors.”
    −­∗ *Clay Kao-Lin*, the Most Interesting Geek in the World­∗

haarg:
It isn't really a problem of incompatibility.  The various make dialects are always different and incompatible.  That is expected.  The problem this PR is meant to solve is building a dist with two different perl builds (if they are using different make versions) without cleaning the source directory.  There are other generated files that will cause issues when attempting this, so it doesn't work in the general case.  I'm pretty certain this case would only ever be used by one person, and would break a number of other things.  And it can already be done with environment variables.

I'm closing this because it would cause bugs and doesn't solve any problems that can't be solved better a different way.

PR #218, opened 2015-04-19
Add sharedir support

Leont:
This patch adds sharedir support to ExtUtils::MakeMaker, so it is no longer necessary to depend on an external module such as File::ShareDir::Install.


karenetheridge:
omg +:100:

rehsack:
It's wrong in the same way as File::ShareDir::Install is (doesn't allow to use share in it's meaning - so it confuses non-ptg people), nor it's backward compatible (but an extension would be).

When you really want support for "immutable data files at same place as code resides" - please rename it to "code_data_dir" or something like that. That makes it much clearer.

Leont:
> It's wrong in the same way as File::ShareDir::Install is (doesn't allow to use share in it's meaning - so it confuses non-ptg people), nor it's backward compatible (but an extension would be).

I assume you mean it's not using /usr/share and such for sharedirs. While I think that is a good goal, fixing it properly will require core support. I don't think long-term dreams like that should get in the way of fixing problems we're facing now.

rehsack:
Not only - I think, packagers, admins and such folks shall be able to specify another root instead of 

    File::Spec->catdir('$(INST_LIB)', qw(auto share ...) )

It would be sane to have it two-staged - one default value for INST_SHARE (as computed above) and the installation itself (EU::I) uses $(INST_SHARE)/...

Further - I don't think an introduction into EU::MM is sane when Perl 5.6 (or 5.8.4) should be supported out-of-the-box. Unless we decide to move to 5.14 (which supports configure_requires out of the box), the EU::MM extension is dangerous - because it breaks such condistions ...

That's why I think, we should try to bring ExtUtils::MakeMaker::Extensions (idea over dedicated implementation) forward and makes it easy to extend EU::MM in a backward portable way.

Leont:
> It would be sane to have it two-staged - one default value for INST_SHARE (as computed above) and the installation itself (EU::I) uses $(INST_SHARE)/...

Yes, that is in line with my core support idea. It could be a good intermediate step.

> Further - I don't think an introduction into EU::MM is sane when Perl 5.6 (or 5.8.4) should be supported out-of-the-box. Unless we decide to move to 5.14 (which supports configure_requires out of the box), the EU::MM extension is dangerous - because it breaks such condistions ...

> That's why I think, we should try to bring ExtUtils::MakeMaker::Extensions (idea over dedicated implementation) forward and makes it easy to extend EU::MM in a backward portable way.

All of that already breaks on such machines now, because File::ShareDir::Install also needs to be configure-required, and so will ExtUtils::MakeMaker::Extensions.

rehsack:
> Am 23.04.2015 um 19:44 schrieb Leon Timmermans <notifications@github.com>:
> 
> It would be sane to have it two-staged - one default value for INST_SHARE (as computed above) and the installation itself (EU::I) uses $(INST_SHARE)/...
> 
> Yes, that is in line with my core support idea. It could be a good intermediate step.

Ok, since this is my strong "disagree".

> Further - I don't think an introduction into EU::MM is sane when Perl 5.6 (or 5.8.4) should be supported out-of-the-box. Unless we decide to move to 5.14 (which supports configure_requires out of the box), the EU::MM extension is dangerous - because it breaks such condistions ...
> 
> That's why I think, we should try to bring ExtUtils::MakeMaker::Extensions (idea over dedicated implementation) forward and makes it easy to extend EU::MM in a backward portable way.
> 
> All of that already breaks on such machines now, because File::ShareDir::Install also needs to be configure-required, and so will ExtUtils::MakeMaker::Extensions.

Maybe - we should probably figure out how we can sane bundle it (either EU::MM or EU::MM::_X) without running into to much trouble - or develop out an easy fallback method.

File::ShareDir::Install (neither File::ConfigDir::Install - and you don't support etcdir, but you might should ...) has much dependencies, so it's very cheap to bundle that. With the proposal of mst for ExtUtils::MakeMaker::Extensions it will be slightly heavier - I didn't start hacking because I wanted a bikeshedding to get it as small as possible in Berlin - probably that should be done via cpan-workers@ now :(

Maybe (off-topic) it could be sane to have META.json being preferred over META.yml in modern toolchains - and have META.yml all unsupported (pre-v2) depends as rundepends in modern authoring tools.

rehsack:
Bump + see post on cpan-workers@ regarding File::ShareDir & File::ConfigDir plans.

haarg:
Here's an experiment I threw together that could serve as a companion to this PR:  https://github.com/haarg/ExtUtils-ShareDir

It's basically a copy of the code in this PR, but packaged as a module to be used in inc/.  This would allow a dist to use this interface and still support systems that don't support configure requires.

Leont:
> Here's an experiment I threw together that could serve as a companion to this PR: https://github.com/haarg/ExtUtils-ShareDir

That's similar to how I [prototyped](https://github.com/Leont/file-sharedir-mm/blob/master/lib/File/ShareDir/MM.pm) this, except I didn't keep that entirely up to date wrt `INST_SHARE`.

PR #222, opened 2015-07-01
Fix Cygwin rebase command by adding pipe

Sophira:
Commit 037a0a1d introduced automatic rebasing for Cygwin, but inadvertently left out a pipe character in the necessary command, causing compilations to fail with errors like:

```
/bin/find /usr/lib/perl5/site_perl/5.14/i686-cygwin-threads-64int/auto/ -xdev -name \*.dll /bin/rebase -sOT -
/bin/find: paths must precede expression: /bin/rebase
Usage: /bin/find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]
Makefile:719: recipe for target 'pure_install' failed
make: *** [pure_install] Error 1
```

This pull request adds the missing pipe character. Please note that I have not directly tested this fix; I have only indirectly tested it by adding the pipe character to one of the generated Makefiles, allowing the module in question to report a successful installation. This will need review and testing to make sure this fix is correct.


Sophira:
I have now tested this by installing EUMM from modified source, and the change does indeed allow the module in question (Net::SSLeay) to report a successful installation.

bingos:
Thanks.

Issue #194, opened 2015-01-05
EU::MM should warn on missing TEST_REQUIRES (and perhaps CONFIGURE_REQUIRES too)


mvuets:
I specified modules required for running a test suite in `TEST_REQUIRES`, but nor _Makefile.PM_, neither `make test` did not warn me on missing prereqs. _Makefile.PM_ warns on missing `PREREQ_PM` though, so I expect `TEST_REQUIRES` behave similarly.

An excerpt from #toolchain:

> haarg: EUMM will warn about missing prereqs in PREREQ_PM and BUILD_REQUIRES
> haarg: but apparently not for TEST_REQUIRES
> mohawk: should it? 
> haarg: yes
> mohawk: and if there's a consensus i'll implement it
> mohawk: time estimated: 6 minutes
> ether: should probably warn about unfulfilled CONFIGURE_REQUIRES too, just for completeness :)
> ether: even though usually an error would happen sooner

haarg:
+1 to warning on missing TEST_REQUIRES and CONFIGURE_REQUIRES

Possibly the message for CONFIGURE_REQUIRES should be different?  Usually, running Makefile.PL would error out in this case.  If it didn't though, the user would probably want to re-run Makefile.PL after installing them.

Leont:
+1

mohawk2:
I'm on this. No time estimate since the consensus is against such.

salewski:
+1

PR #227, opened 2015-07-14
Warn on missing prereqs for *_REQUIRES, issue #194.

thrig:
Dunno if anything else need be done, or ramifications of such.


karenetheridge:
Awesome :)

mohawk2:
Looks like a worth addition! Now if only there were a test included... :-)

thrig:
So these just check that C_R and T_R emit messages, and not the full monty of "" version number checks and such, but in the code it looks like everything gets shoved into that one loop... (adding *what* _REQUIRES failed to the message would be more work).

PR #228, opened 2015-08-03
Remove pure_nolink change

mohawk2:
Under GNU make 3.8* and 4.1, perl build fails with -j4. With the pure_nolink (etc) change reverted, it succeeds again. Passes tests on Linux and Win32 "in space".

PR-ing for visibility.

### DIFF
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -1036,11 +1036,7 @@ sub dynamic {
 
     my($self) = shift;
     '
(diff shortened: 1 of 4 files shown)


Leont:
> Under GNU make 3.8* and 4.1, perl build fails with -j4. With the pure_nolink (etc) change reverted, it succeeds again. Passes tests on Linux and Win32 "in space".

Can you explain the failure mode? What fails? Why?

mohawk2:
https://savannah.gnu.org/bugs/?44742

PR #234, opened 2015-09-04
account better for bad version numbers

haarg:
Older versions of EUMM just did numeric comparisons of versions with
warnings disabled.  Newer versions use version objects to do the
comparisons.  The older checks allowed some malformed versions to pass
their checks in some cases that the version objects reject.  Also, the
version listed in the files is expected to be a valid lax version, and
would crash if this wasn't the case.

Instead of rejecting these versions or crashing, do some extra massaging
to pick out a numeric prefix of a malformed version and use that.  This
should mostly match the old behavior for versions in this form, but
we're now issuing warnings about all of these cases.

Fixes [RT#106873](https://rt.cpan.org/Ticket/Display.html?id=106873)

### DIFF
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -451,8 +451,11 @@ sub new {
                 $self->{$key},
                 {
                   bad_version_hook => sub {
-                    carp "Unparsable version '$_[0]' for prerequisite $_[1] treated as 0";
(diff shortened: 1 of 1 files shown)


karenetheridge:
problem noted in the 7.06 regressions list - https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/235

Issue #235, opened 2015-09-04
informational ticket: 7.06 regressions


karenetheridge:
put links to known 7.06 regressions here.

karenetheridge:
non-lax $VERSIONs are now rejected when checking if prereqs are installed:  https://rt.cpan.org/Ticket/Display.html?id=106873  -- and fixed here -- https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/234

karenetheridge:
11:24 <@rjbs> ExtUtils::MakeMaker 7.04, Unicode::GCString installs fine.
11:25 <@rjbs> With 7.06: https://gist.github.com/rjbs/e5a3566e4c747d69ddbe
11:31 <@rjbs> Worked in 7.05_04.  Does ont work in 7.05_05.
11:32 <@haarg> https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/8e4059c6d6cc1bc9e12fdb98a9538e14de2ffd44
11:32 <+dipsy> [ xs_obj_opt method adds -o to all $(CCCMD) commands to put .o next to .c · Perl-Toolchain-Gang/ExtUtils-MakeMaker@8e4059c · GitHub ] 
11:33 <@haarg> pretty sure that's the commit that broke it
11:33 <@rjbs> certainly seems like it would be
11:39 <@BinGOs> that's the one that broke File::Libmagic too.

karenetheridge:
https://rt.cpan.org/Ticket/Display.html?id=106843 -- Makefile format changed -- fix in the dist, not EUMM

karenetheridge:
Changes in PASSTHRU -- many distribution failures documented here -- https://rt.cpan.org/Ticket/Display.html?id=106808

Leont:
I'm in favor of a quick regression fix release, and leave new developments for 7.10

ribasushi:
Another meta-regression, arguably with lesser priority as it affects dev workflows only, not full-on installations: https://rt.cpan.org/Ticket/Display.html?id=106927. Nevertheless filing it, as ramifications for build-systems relying on `-e blib` *might* be severe.

Leont:
> With 7.06: https://gist.github.com/rjbs/e5a3566e4c747d69ddbe

Essentially, it's already trying to work around the lack of -o. This is all such a fucking mess! :-(

andk:
[...]

PR #238, opened 2015-09-10
Use same build-info dir as Module::Build

mohawk2:
So far as I know, this does not clash with Module::Build or any other toolchain module. I await constructive feedback.

This will address https://rt.cpan.org/Ticket/Display.html?id=106927

cf #235

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,4 @@ MYMETA.*
 ExtUtils-MakeMaker-*
 inc/
 .*.swp
-/_eumm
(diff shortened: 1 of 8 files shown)


shadowcat-mst:
Many existing tools assume that the existence of _build indicates a Module::Build based build process;  as such, this PR will cause much downstream breakage.

Please file an independent issue indicating what genuine problems the _eumm concept will solve so we can determine if it's worth the cost.

PR #240, opened 2015-09-15
- stop makeaperl from polluting @ARGV in cases where ARGV contains ar…

plobbes:
…gs with spaces (ex: perl Makefile.PL LIBS='-L... -l...')

Background:
I ran into this issue when building Net::SSLeay and found that it ends up running through MakeMaker twice via inc::Module::Install code and on the second pass through @ARGV was polluted (extra single quotes wrapping args with spaces (LIBS in my case since we have a custom build of OpenSSL in a custom location/run path).

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -2483,8 +2483,8 @@ $(MAKE_APERL_FILE) : static $(FIRST_MAKEFILE) pm_to_blib
 
 	foreach (@ARGV){
 		my $arg = $_; # avoid lvalue aliasing
-		if ( $arg =~ /\s/ ) {
-			$arg =~ s/=(.*)/='$1'/;
+		if ( $arg =~ /(^.*?=)(.*['\s].*)/ ) {
+			$arg = $1 . $self->quote_literal($2);
 		}
 		push @m, " \\\n\t\t$arg";
 	}
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -2482,12 +2482,12 @@ $(MAKE_APERL_FILE) : static $(FIRST_MAKEFILE) pm_to_blib
 		MAKEAPERL=1 NORECURS=1 CCCDLFLAGS=};
 
 	foreach (@ARGV){
-		if( /\s/ ){
-			s/=(.*)/='$1'/;
+		my $arg = $_; # avoid lvalue aliasing
+		if ( $arg =~ /\s/ ) {
+			$arg =~ s/=(.*)/='$1'/;
 		}
-		push @m, " \\\n\t\t$_";
+		push @m, " \\\n\t\t$arg";
 	}
-#	push @m, map( " \\\n\t\t$_", @ARGV );
 	push @m, "\n";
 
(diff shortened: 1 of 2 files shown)


Leont:
Looks sensible to me

t/MM_Unix.t
```diff
@@ -222,3 +222,13 @@ foreach (qw/ EXPORT_LIST PERL_ARCHIVE PERL_ARCHIVE_AFTER /)
     like( $t->{CCFLAGS}, qr/\-DMY_THING/,    'cflags retains CCFLAGS' );
 }
 
+#
+{
+    my @orig = @ARGV;
```
mohawk2: Adding tests is awesome! Think this would be better as `local @ARGV = @ARGV`.

Leont:
~~~~
local @ARGV = ("var=val with spaces");
~~~~

mohawk2:
A further thought: it will probably best to handle (and also test for) ARGV values that have quotes in; consider a `-I` argument for a place-with-a-space (cf Windows).

lib/ExtUtils/MM_Unix.pm
```diff
@@ -2482,12 +2482,12 @@ $(MAKE_APERL_FILE) : static $(FIRST_MAKEFILE) pm_to_blib
 		MAKEAPERL=1 NORECURS=1 CCCDLFLAGS=};
 
 	foreach (@ARGV){
```
mohawk2: Now I think of it: `foreach my $arg (@ARGV) ...`

Leont: The whole point is not to do lvalue aliasing ;-)

plobbes: > A further thought: it will probably best to handle (and also test for) ARGV values that have quotes in...

Agreed. However, I would consider the handling embedded quotes as a separate issue (to be fixed in a separate patch with more tests).  The fix here is specifically to avoid the ARGV pollution which should never happen.

> Now I think of it: foreach my $arg (@ARGV) ...

@Leont hinted at this, but to be clear, if you do this you're still getting an alias to the value within the array.  It's an interesting and incredibly subtle feature of Perl which makes it non-obvious that what you suggest above doesn't work as you're probably expecting.  Feel free to test/try that though to see this in action.

mohawk2: Just did. Every day is a school day.

plobbes:
Are we OK with the fix as-is now?  Or were we hoping to also pull in a fix for handling embedded quotes within ARGV?  (FWIW: My vote is to keep that as a separate issue/fix.)

Leont:
>  Or were we hoping to also pull in a fix for handling embedded quotes within ARGV?

We already have utility functions for this. I suspect it's just a matter of using something like
~~~~
$arg =~ s/ (?<= =) (.*) /$self->quote_literal($1)/xse;
~~~~

That s modifier fixes newlines in arguments ;-)

plobbes:
Hopefully this latest change addresses the concerns about embedded single quotes in a simple enough way (while still not polluting ARGV of course).

mohawk2:
Looks good to me. We're not testing the single-quote-protection yet, though.

PR #244, opened 2015-11-08
Remove _eumm dir

haarg:
This reverts the changes adding the _eumm or blib/_eumm directory.

blib is meant to be created by the pm_to_blib make target, so creating it during configure leads to issues.

Having an _eumm directory in the top level is what we want long term, but adding files like that will lead to them being unintentionally shipped with many dists since _eumm won't be listed in MANIFEST.SKIP files normally.

We could possibly add additional encoding to deal with part of what _eumm is meant to fix.  But this PR leaves us no worse off than the current stable release, and should fix the last known regression listed in #235.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ MYMETA.*
 ExtUtils-MakeMaker-*
 inc/
 .*.swp
-/_eumm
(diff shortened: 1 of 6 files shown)


Leont:
I think this is the only possible way forward.

karenetheridge:
+1

mohawk2:
+1 from me. Was actually on the verge of putting together the tuits to do it myself. Nice work @haarg !

PR #254, opened 2016-02-14
Remove private variables $Updir, $Rootdir when not used

dolmen:
A minor change removing unused variables.


lib/ExtUtils/MM_Any.pm
```diff
@@ -18,8 +18,7 @@ use ExtUtils::MakeMaker::Config;
 # So we don't have to keep calling the methods over and over again,
 # we have these globals to cache the values.  Faster and shrtr.
 my $Curdir  = __PACKAGE__->curdir;
-my $Rootdir = __PACKAGE__->rootdir;
-my $Updir   = __PACKAGE__->updir;
+#my $Updir   = __PACKAGE__->updir;
```
mohawk2: If we're removing them, let's remove them entirely not just comment?

dolmen: $Updir has a risk of reappearing.
$Rootdir, not.

PR #260, opened 2016-03-19
Version range warning

### LINKED ISSUE #215: need to add a configure_requires entry for self at 7.05_01 if a version range is seen
I used a range here: https://metacpan.org/source/ETHER/Dist-Zilla-Plugin-PromptIfStale-0.041/Makefile.PL#L24

In earlier EUMMs the version range is treated as 0, rather than respecting the stated prereq.

mohawk2:
Fix #215.


lib/ExtUtils/MakeMaker.pm
```diff
@@ -427,6 +427,22 @@ sub _has_cpan_meta_requirements {
     };
 }
 
+sub _min_EUMM {
+    my ($self) = @_;
+    my @eumm_versions = sort grep defined, map {
+        $_->{'ExtUtils::MakeMaker'}
+    } grep defined, @{$self}{@PREREQ_KEYS};
```
haarg: CONFIGURE_REQUIRES is the only place that is useful for the place this is being used, so checking all of the phases is incorrect.  It's also possible to provide the configure requires only via META_ADD/META_MERGE, so this could result in spurious warnings.

lib/ExtUtils/MakeMaker.pm
```diff
@@ -609,7 +628,7 @@ END
         elsif (
             $cmr
                 ? !$cmr->accepts_module($prereq, $pr_version)
-                : $required_version > $pr_version
+                : $required_version lt $pr_version
```
haarg: This reverses the conditional, breaking it.  The switch from numeric to stringy comparison is also unexplained in the commit message or anywhere else.  And I'm pretty sure we want to stick with numeric comparisons here.

lib/ExtUtils/MakeMaker.pm
```diff
@@ -427,6 +427,22 @@ sub _has_cpan_meta_requirements {
...
 }
 
+sub _min_EUMM {
+    my ($self) = @_;
+    my @eumm_versions = sort grep defined, map {
+        $_->{'ExtUtils::MakeMaker'}
+    } grep defined, @{$self}{qw(CONFIGURE_REQUIRES)};
+    return 0 unless @eumm_versions;
+    $eumm_versions[0];
+}
```
haarg: `$self->{CONFIGURE_REQUIRES} && $self->{CONFIGURE_REQUIRES}{'ExtUtils::MakeMaker'} || 0`

haarg:
Based on discussion on IRC, I think we want to drop this idea.

Leont:
Yes we do

mohawk2:
If we do actually do something like this, then:
* @Leont is right there needs to be a bit more infrastructure
* @haarg 's idea of doing it in / after META has been calculated is a great one as then one only need check for one thing
* @karenetheridge is right that we'd also check for newer features like TEST_REQUIRES, arrayref-AUTHOR, etc

karenetheridge:
I'm very sorry for stirring up a hornet's nest. However it's good to know this nest is here so we don't step in it again, and maybe have some ideas for tidying it up in the future :)

haarg:
Another problem with this idea is that many Makefile.PL files will adjust how they call EUMM based on the EUMM version.  So using a newer EUMM option without listing it in configure requires is perfectly fine.  Adding these warnings would show false positives to users, and we can't detect users vs authors.

Arrayref AUTHOR doesn't make sense to warn about.  AUTHOR isn't really relevant on the user side, and if it is "broken" it doesn't impact installation.

You are welcome to keep working on this if you want, but I don't think it's possible to do in a way that I will find acceptable.

PR #263, opened 2016-04-23
lazy load Time::HiRes in ExtUtils::Command::MM

haarg:


lib/ExtUtils/Command/MM.pm
```diff
@@ -15,10 +15,14 @@ $VERSION = eval $VERSION;
 
 my $Is_VMS = $^O eq 'VMS';
 
-eval {  require Time::HiRes; die unless Time::HiRes->can("stat"); };
-*mtime = $@ ?
- sub { [             stat($_[0])]->[9] } :
- sub { [Time::HiRes::stat($_[0])]->[9] } ;
+sub mtime {
+  no warnings 'redefine';
+  *mtime = (eval { require Time::HiRes } && defined &Time::HiRes::stat)
```
ribasushi: @haarg probably a good idea to guard $@ from this eval: who knows where mtime will be called...

PR #267, opened 2016-04-27
Fix regression with small fractional numeric versions

bingos:
If a version number is passed in numeric form, C::M::R will use its stringified form for all its processing.
This doesn't play well with small fractional versions like 0.000005, which will be stringified as '5e-06',
and thus rejected as invalid.

EUMM then proceeded misparse the rejected number, so in the above case it became '5'

Also added test-case to vstrings.t

### DIFF
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -452,9 +452,14 @@ sub new {
                 {
                   bad_version_hook => sub {
                     #no warnings 'numeric'; # module doesn't use warnings
-                    my ($fallback) = $_[0] ? ($_[0] =~ /^([0-9.]+)/) : 0;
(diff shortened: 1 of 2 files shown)


haarg:
Looks good to me.

mohawk2:
And me.

Leont:
It wouldn't deal well with numbers that have more than 6 digits, not that those are common

haarg:
Can version.pm handle these version correctly?  If so, it would probably be better to just let it do so.

Otherwise maybe we need to sprintf '%.12f' =~ s/0+$//

Leont:
~~~
$ perl -e 'printf "%f\n", 5e-7'
0.000000
$ perl -e 'printf "%s\n", version->new(5e-7)'
0.0000005
~~~

PR #272, opened 2016-06-14
prevent EUMM::Locale from warning with old Win32.pm

haarg:
Older DynaLoader inherits from AutoLoader, so nonexistent function calls
will trigger AUTOLOAD, but with a warning over deprecated behavior.

With an older Win32.pm that doesn't have the functions we are looking
for, checking for their existence with a call in an eval will trigger
this warning.  That warning breaks our tests.

Replacing the eval+call with a defined check on the function is a more
appropriate way to check for functions, and will avoid the warnings.

This unfortunately means we diverge from Encode::Locale.  Hopefully the
next version of that module will include this fix.


karenetheridge:
(patched in https://github.com/gisle/encode-locale/pull/19)

Issue #274, opened 2016-07-01
EU::MM 7.12 breaks CSS::Sass


mgreter:
Hello

This was originally reported at https://github.com/sass/perl-libsass/issues/25

I have tried to debug it but have hit some other obstacles on the road. Therefore I'm positing this here, in the hope somebody can give me more hints how to address this properly. I also saw that about 3 tickets on the [cpan bug tracker] [1] reported something similar, but unsure if they are directly related.

The reason it started to fail is the difference in the following two lines:
```bash
x86_64-linux-gnu-gcc -shared -L/usr/local/lib -fstack-protector-strong libsass/src/*.o Sass.o -lstdc++ -o blib/arch/auto/CSS/Sass/Sass.so # EU::MM 7.10
x86_64-linux-gnu-gcc -shared -L/usr/local/lib -fstack-protector-strong -o blib/arch/auto/CSS/Sass/Sass.so -lstdc++ libsass/src/*.o Sass.o # EU::MM 7.12
```
The problem is really seems to be the order of `-lstdc++` vs  `libsass/src/*.o`

I'm not sure why these should produce different results, but they unfortunately do. My tests indicate that it works if the `-lstdc++` option comes after at least the first object file, but it fails if it is added before. 

The `-lstdc++` flag is directly from ExtUtils::CppGuess (`{'dynamic_lib'}->{'OTHERLDFLAGS'}`). I saw from the generated Makefile that this flag would work if it would have been in `$LDLOADLIBS` or `$INST_DYNAMIC_FIX`. Next I tried to move the detected `OTHERLDFLAGS` directly to `LIBS`, but then I get `Warning (mostly harmless): No library found for -lstdc++`.

After a bit more debugging I found that my `libpth` (from Config) seems wrong:
```
# perl -e 'use Config; use Data::Dumper; print Dumper(\%Config)' | grep "'\(libspath\|libpth\|incpth\)"
'incpth' => '/usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/include /usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/include-fixed /usr/include',
'libpth' => '/usr/local/lib64 /lib64 /usr/lib64 /usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/include-fixed /usr/lib /lib/../lib64 /usr/lib/../lib64 /lib',
[...]

PR #275, opened 2016-07-03
Fix ordering issue in OTHERLDFLAGS  #274

Leont:
Ordering of linker arguments matters a lot, the object file should
be given after the libraries that are being linked to.


mohawk2:
There is often a need to put the arguments in a particular order. Unfortunately, that order can vary: http://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc . What matters is to put the objects before the libraries. This change does not affect that, only moves the `-o output` around.

I also think making changes to the behaviour of this sort of thing without adding tests is a large, unnecessary risk.

Leont:
> There is often a need to put the arguments in a particular order. Unfortunately, that order can vary: http://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc .

My description was wrong (updated now), but the effect was exactly right.

> What matters is to put the objects before the libraries. This change does not affect that, only moves the -o output around.

It switches the order back to 7.10's Objects, OtherLDFLags, Output, whereas in 7.12 it's Output OtherDLFlags, Objects

> I also think making changes to the behaviour of this sort of thing without adding tests is a large, unnecessary risk.

Then why didn't you add those tests before changing the order of the arguments? I didn't introduce anything new, I merely restored the old order.

mohawk2:
No-one (including me, and presumably you) spotted I was changing that order. So while this change may fix a bug, there is no guarantee that it will stay. You are making a patch without making a test for that patch.

Leont:
If you think it's so important, why don't you write those tests? I would be a lot more constructive than complaining about me fixing bugs.

PR #279, opened 2016-08-28
Make perllocal.pod files reproducible

jmdh:
As per https://bugs.debian.org/834190 and https://bugs.debian.org/835815 this patch makes the perllocal.pod files generated by ExtUtils::MakeMaker be reproducible across builds. This patch is part of the Reproducible Builds[0](https://reproducible-builds.org/) project:

### DIFF
--- a/lib/ExtUtils/Command/MM.pm
+++ b/lib/ExtUtils/Command/MM.pm
@@ -219,7 +219,8 @@ sub perllocal_install {
                            : @ARGV;
 
     my $pod;
-    $pod = sprintf <<'POD', scalar(localtime), $type, $name, $name;
+    my $time = gmtime($ENV{SOURCE_DATE_EPOCH} || time);
+    $pod = sprintf <<'POD', scalar($time), $type, $name, $name;
  =head2 %s: C<%s> L<%s|%s>
 
  =over 4
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -3869,7 +3869,7 @@ Obsolete, deprecated method. Not used since Version 5.21.
 sub writedoc {
 # --- perllocal.pod section ---
     my($self,$what,$name,@attribs)=@_;
-    my $time = localtime;
+    my $time = gmtime($ENV{SOURCE_DATE_EPOCH} || time);
     print "=head2 $time: $what C<$name>\n\n=over 4\n\n=item *\n\n";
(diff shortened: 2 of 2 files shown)


Leont:
I'm not sure this is quite the right fix. I guess we want `gmtime` with `SOURCE_DATE_EPOCH` but `localtime` with `time`.

jmdh:
We have to use gmtime to make it reproducible across different environments, and changing the implicit timezone depending on whether it's overriden by an environment variable seems equally wrong. At least I assume that's the rationale behind this pattern; @lamby?

lamby:
No real rationate apart from most people don't care about the localtime/gmtime on a "day" level...

Feel free to split it like (untested):

```
if ($ENV{SOURCE_DATE_EPOCH}) {
    my $time = gmtime($ENV{SOURCE_DATE_EPOCH});
} else {
    my $time = localtime;
}
$pod = sprintf <<'POD', scalar($time), $type, $name, $name;
```

(etc.)

dolmen:
There are two things changed by this patch:
- the timestamp used ($ENV{SOURCE_DATE_EPOCH} instead of system time)
- how that timestamp is formated: localtime vs gmtime

I'm ok with the first one.

About the second one, how do you expect the build to be reproducible if the locale is not fixed in the build environment? Is the Debian reproducible build environment locale-dependent?

jmdh:
Yes, the recommendation is that builds use LC_ALL=C: https://reproducible-builds.org/docs/locales/

lamby:
> is the Debian reproducible build environment locale-dependent

It is locale-dependent.

However, the ``gmtime`` vs ``localtime`` is not a locale issue, it's a timezone issue and is therefore orthogonal to any advice regarding ``LC_*`` environment variables.

Leont:
Actually, the whole concept of perllocal seems rather non-reproducible to me

dolmen:
Too bad this has been merged. A reproducible build should just disable `perllocal` generation (`make pure_inst_perl`).
This patch only brings Debian specific bits.

lamby:
@dolmen wrote:

>  A reproducible build should just disable perllocal generation

How would you suggest we do that reliably? Can we adjust this so that if ``SOURCE_DATE_EPOCH`` is set we just skip these bits "automatically" or are you saying we should call a different target (as *well* as export ``SOURCE_DATE_EPOCH`` for other time-related stuff)…?  My gut/experience tells me the former would be more useful/practical.

dolmen:
> How would you suggest we do that reliably?

Written above: `make pure_inst_perl`

PR #280, opened 2016-10-03
META_ADD/MERGE default meta version based on each other

haarg:
Prior to commit 91b7ff8b56d529cfcee575b7858878ea0d415a0d (released in
v7.05_13), META_ADD and META_MERGE will be erroneously mixed together
when processing.  This essentially means that if a meta-spec version is
specified in one, it will be used for the other as well.

Now that META_ADD and META_MERGE are properly kept separate, they can
specify data in different spec versions.  But if a version is only
specified in one, it makes sense to use it for the other as well.  This
also matches the old behavior, even if it was at the time unintentional.

### DIFF
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -1319,7 +1319,8 @@ sub metafile_data {
     my $v1_add = _metaspec_version($meta_add) !~ /^2/;
 
     for my $frag ($meta_add, $meta_merge) {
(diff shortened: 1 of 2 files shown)


karenetheridge:
I worked around this issue in https://github.com/p5sagit/JSON-MaybeXS/commit/da4884da1ec1d4c3690350118ebb998f2d81a11c and https://github.com/p5sagit/JSON-MaybeXS/commit/2c3375e8110fa8feec9c64eac6e9c3a2901df78f.

Issue #283, opened 2017-01-21
[PATCH] Make EU::MM set PERL_UNSAFE_INC. (CVE-2016-1238)


toddr:
Opening this ticket to discuss work required to support a perl without . in @INC by default.

See also https://rt.perl.org/Ticket/Display.html?id=130467 and https://rt.perl.org/Ticket/Display.html?id=127810 for core perl discussions.

My doc is here http://blogs.perl.org/users/todd_rinaldo/2016/11/how-removing-from-inc-is-about-to-break-cpan.html

I'll submit a pull request shortly for discussion.

toddr:
I've amended the patch to not set the PERL_USE_UNSAFE_INC environment variable if it's already set.

toddr:
For the time being, I'm cancelling this issue. It appears that solving the majority of issues with Test::Harness should make the remaining failures manageable.

dolmen:
Using a CVE ref in the title is misleading. Especially because this patch on EUMM aims to preserve the current behavior (for backwards compatibility), not to fix it.

toddr:
> Using a CVE ref in the title is misleading. Especially because this patch on EUMM aims to preserve the current behavior (for backwards compatibility), not to fix it.

The point was to give links to the background information on why it was being pursued. I had no intention of implying this fixes a CVE. You're right the subject should probably be changed.

PR #284, opened 2017-01-21
Add PERL_USE_UNSAFE_INC=1 during Makefile.PL and make test

toddr:
https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/283

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -3702,6 +3702,10 @@ EOF
     join "", @m;
 }
 
+sub _unsafe_inc_string {
+    return exists $ENV{PERL_USE_UNSAFE_INC} ? '' : 'PERL_USE_UNSAFE_INC=1 ';
+}
+
 =item test_via_harness (override)
 
 For some reason which I forget, Unix machines like to have
@@ -3711,7 +3715,8 @@ PERL_DL_NONLAZY set for tests.
 
 sub test_via_harness {
     my($self, $perl, $tests) = @_;
-    return $self->SUPER::test_via_harness("PERL_DL_NONLAZY=1 $perl", $tests);
+
+    return $self->SUPER::test_via_harness("PERL_DL_NONLAZY=1 " . _unsafe_inc_string() . $perl, $tests);
 }
 
 =item test_via_script (override)
@@ -3722,7 +3727,7 @@ Again, the PERL_DL_NONLAZY thing.
 
 sub test_via_script {
     my($self, $perl, $script) = @_;
-    return $self->SUPER::test_via_script("PERL_DL_NONLAZY=1 $perl", $script);
+    return $self->SUPER::test_via_script("PERL_DL_NONLAZY=1 " . _unsafe_inc_string() . $perl, $script);
 }
 
 
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -5,6 +5,11 @@ use strict;
 
 BEGIN {require 5.006;}
 
+# Assure anything called from Makefile.PL includes . in @INC.
+BEGIN {
+    $ENV{PERL_USE_UNSAFE_INC} = 1 unless exists $ENV{PERL_USE_UNSAFE_INC};
+}
+
 require Exporter;
 use ExtUtils::MakeMaker::Config;
 use ExtUtils::MakeMaker::version; # ensure we always have our fake version.pm


haarg:
This doesn't fix the issue on Win32.

ExtUtils::MakeMaker is used in places other than Makefile.PL.  Setting the env var unconditionally on load seems like a bad idea.  I'm also not sure what that will actually fix.  It doesn't help the current process, which is where it is most likely to be needed.  It would only help if Makefile.PL is running additional perl processes, which doesn't seem very common.

If possible, we should allow a user to set PERL_USE_UNSAFE_INC=0 and then not override that choice.

Leont:
> This doesn't fix the issue on Win32.

Moving to MM_Any would be the easy solution, but odds are it would break VMS (I have no idea).

> ExtUtils::MakeMaker is used in places other than Makefile.PL. Setting the env var unconditionally on load seems like a bad idea. I'm also not sure what that will actually fix. It doesn't help the current process, which is where it is most likely to be needed. It would only help if Makefile.PL is running additional perl processes, which doesn't seem very common.

Agreed on both counts.

> If possible, we should allow a user to set PERL_USE_UNSAFE_INC=0 and then not override that choice.

Yeah we should do that consistently wherever we set it.

toddr:
@haarg:
"This doesn't fix the issue on Win32."
I see no where else in the code base that sets PERL_DL_NONLAZY=1 Are you certain this doesn't fix win32? How does win32 do it? I'm open to a counter patch. win32 could also be a follow on patch.

toddr:
> ExtUtils::MakeMaker is used in places other than Makefile.PL. Setting the env var unconditionally on load seems like a bad idea. I'm also not sure what that will actually fix. 
This addresses Makefile.PL invoked from somewhere other than a CPAN client. There are 2 goals which we're trying to address:
1. Help out funky Makefile.PL wierdness where something happens after use ExtUtils::MakeMaker. Any perl code Makefile.PL called would then be sure to get . in @INC.
2. make it so make test always has . in @INC.

To be honest the second goal is the only one I feel strongly about. The hundred CPAN modules that 1 will cover over, I'm happy to open RT tickets about and fix the module's assumptions.

I've been trying to engage a conversation on this in several forums for as many months. If you have an alternative suggestion I would be very happy to hear your thoughts.

> If possible, we should allow a user to set PERL_USE_UNSAFE_INC=0 and then not override that choice.
The exists check is a good idea.

haarg:
Setting `PERL_USE_UNSAFE_INC` in Makefile.PL (via ExtUtils::MakeMaker) won't cause the current process to have . in `@INC`.  It would only fix cases where perl was being called as a sub process in Makefile.PL.  There aren't going to be many cases that would fix.  Adding . to `@INC` would fix more things, but I still don't like it given that EUMM is used in other contexts.

PERL_DL_NONLAZY isn't set by EUMM on Win32.  And from what I can tell, perl doesn't even look at it anywhere but HP-UX and Symbian.  On Win32, the MM_Any implementation is preferred over MM_Unix.  `ENVVAR=val command` is a bourne shell feature.  It won't work in `cmd.exe`.  A possible alternative is `set ENVVAR=val & command`.  Other non-unix systems will need their own handling.

Unfortunately doing an exists check won't really be possible on a per make invocation basis.  But we could allow its setting at Makefile.PL time to control the rest.

jmdh:
I'm afraid I don't have exact numbers, but when we made this change
in Debian a patch similar to this one was applied and fixed substantial
numbers of packages where a child perl process was being invoked from
Makefile.PL. My estimation[1] is that this change alone fixed at least 200 of
the original 800 failing ones. Combined with setting our packaging helpers
to invoke the top-level Makefile.PL with -I. fixed the majority of build
failures. So there is definite value in this patch.

Patches to the CPAN client programs, which are often responsible for
invoking Makefile.PL, are already planned (or submitted, I'm not sure).

[1] My actual notes say that 808 packages failed after we removed '.'
globally, and after adding the Extutils-Makemaker patch, an alphabetical
run which was interrupted at 'libs[...]' had 315 failures.

haarg:
@jmdh do you have any specific examples of failures that were fixed by that change?  Looking at what people have done might help evaluate solutions.

jmdh:
On Mon, Jan 23, 2017 at 07:11:13PM -0800, Graham Knop wrote:
> @jmdh do you have any specific examples of failures that were fixed by that change?  Looking at what people have done might help evaluate solutions.

Apologies, those numbers included the result of invoking Makefile.PL
with -I. too. The actual number of Extutils::MakeMaker patches fixed
is much smaller. Because the analysis was done with several changes at
once it's quite time-consuming to dig through the logs, but here are some
examples from 'libs and 'libt':

http://www.larted.org.uk/~dom/tmp/makemaker-inc-logs/

Dominic.

Leont:
> http://www.larted.org.uk/~dom/tmp/makemaker-inc-logs/

All of these look like test-time failures to me.

jmdh:
In that case I think my evidence is invalid, sorry :( It's been such
a long time since we did all this, lots of different fixes were needed
for different things but I don't have the details noted down properly.

@toddr do you have any better examples based on your work?

toddr:
> In that case I think my evidence is invalid, sorry :( It's been such
> a long time since we did all this, lots of different fixes were needed
> for different things but I don't have the details noted down properly.

My response is the same. it's been even longer for me. As I understand things, toolchain gang is on the hook for the next action. we're waiting for their response on what they need to proceed.

Leont:
> We discussed this a little recently on #p5p. I can deluge them with build failures but my question to them was what then?

I think step one is fixing the test time failures, because those can be fixed relatively painlessly. Then see what fails other than Module::Install, and see if we can fix that on this side of the toolchain or not.

toddr:
> I think step one is fixing the test time failures, because those can be fixed relatively painlessly. Then see what fails other than Module::Install, and see if we can fix that on this side of the toolchain or not.

Ok. That's this patch isn't it?

> If not the CPAN client patches are the only thing we have.

Even the CPAN client authors are asking for the same list of failures as you guys have. I get that this is your procedure. But I need to know what evidence you require to merge this.

Alternatively If this patch is not acceptable to you guys, I'm good with that. But we need an alternative plan.

Are we going whack-a-mole on this problem?

haarg:
This patch does not properly address the test time failures.  It doesn't address non-Unix platforms, most notably Win32.  Test::Harness is most likely going to be a better place to do that.  We don't need more evidence for that, but the work needs to be done.  I believe @Leont has a start on this.

Setting the env var when EUMM is loaded is unacceptable.  It would mean the env var would be set for any program that uses IPC::Cmd::can_run.  I was asking for evidence because it isn't obvious to me that it would even fix anything.  If it does indeed fix things, I'd like to see what the actual failures are so we can come up with an acceptable solution.  What seems more likely to fix things is adding `.` to `@INC` when EUMM is loaded, but I also don't think that is an acceptable change.

Having the variable set at build time is also unaddressed by this PR.  Unfortunately, it won't have something like Test::Harness that will provide an easier place to set the environment variable.

Cross linking andk/cpanpm#108 miyagawa/cpanminus#521

toddr:
Test::Harness is insufficient. I know this because that was my original strategy when I addressed this in my distribution. 

Worse, you'll end up with the environment variable being set in places it's not ideal to do so.  The goal is to set it only during make / make test / make install. All other times, you want it to fail (for instance when prove is run). 

I will provide a list of modules where Test::Harness does not fix the problem.

Todd

haarg:
@toddr Please also provide the patch you tried to use with Test::Harness.

haarg:
Note that the test_harness sub called in the Makefile is actually from ExtUtils::Command::MM, so if we wanted this change to only effect `make test`, we could easily set PERL_USE_UNSAFE_INC there rather than in Test::Harness.

haarg:
This was handled in Test::Harness instead, so I'm going to close this.

Issue #91, opened 2014-03-02
Our inheritance is screwed up


Leont:
We have a historical split of most of our functionality between MM_Any and MM_Unix, the latter being a subclass of the former. However, all currently used MM_\* classes are subclasses of MM_Unix, which suggests that this distinction is obsolete (MM_MacOS was an example of a direct MM_Any subclass).

To make things more complicated, various subclasses for non-unix operating systems(Win32, VMS, DOS, BeOS) are `@ISA = (MM_Any, MM_Unix)`, which is a rather peculiar order. AFAIK that only matter for 6 methods: all_target, init_platform, init_tools, platform_constants, test_via_harness, test_via_script.

I suspect we could move all of MM_Unix into MM_Any except those methods, and would end up with a simpler system. YMMV.

mohawk2:
Note to whomever does this: File::Find returns paths /-separated, while File::Spec obviously is FS-dependent. This can cause mismatches in make targets.

mohawk2:
Another issue is EU::Liblist[::Kid] are a bit "challenging", and the code should instead be incorporated into MM_*.

Leont:
> Another issue is EU::Liblist[::Kid] are a bit "challenging", and the code should instead be incorporated into MM_*.

Yeah, that code is just really confused if you ask me. I can't imagine why it is the way it is.

Leont:
Come to think about this, this may break a few Makefile.PLs that peek/poke into MM_Unix. A cpan grep suggests this is mainly people calling ExtUtils::MM_Unix::parse_version directly, so if we explicitly import that from MM_Any to MM_Unix

This may affect [HPUX-SDUX](https://metacpan.org/source/ASARIH/HPUX-SDUX-0.03/SDUX.pm#L228) (failure to use `SUPER::`), [DBD::Solid](https://metacpan.org/source/JSLAGEL/DBD-Solid-0.20a/Makefile.PL#L128) (another failure to use `SUPER::`) and [Test::File](https://metacpan.org/source/BDFOY/Test-File-1.44/Makefile.PL#L1) (failure to use `@Overridable`).

rurban:
FYI: The current inheritance is broken with c3 for MM_Win32 and MM_VMS.

rurban:
FYI: I started working on that in my c3 branch. 
use mro 'dfs' as fallback is not possible because mro is a XS and EUMM needs to be run with miniperl in CORE.

[...]

PR #286, opened 2017-02-13
Fix multiple inheritance for c3

rurban:
With multiple inheritance (@ISA>1), the c3 mro
refuses to load the most specific after the least specific base class.
The general rule to avoid inconsistency for @ISA is
from most specific to least specific.

cperl switched to c3 already (perl5 might do later evtl.)
and fails with "Inconsistent hierarchy during C3 merge".

(The other culprit is usually Exporter at the front of an @ISA btw).

### DIFF
--- a/lib/ExtUtils/MM_VMS.pm
+++ b/lib/ExtUtils/MM_VMS.pm
@@ -15,12 +15,12 @@ BEGIN {
 
 use File::Basename;
 
-our $VERSION = '7.25_01';
+our $VERSION = '7.25_02';
 $VERSION = eval $VERSION;
 
 require ExtUtils::MM_Any;
 require ExtUtils::MM_Unix;
-our @ISA = qw( ExtUtils::MM_Any ExtUtils::MM_Unix );
+our @ISA = qw( ExtUtils::MM_Unix ExtUtils::MM_Any );
 
 use ExtUtils::MakeMaker qw($Verbose neatvalue _sprintf562);
 our $Revision = $ExtUtils::MakeMaker::Revision;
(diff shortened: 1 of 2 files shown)


Leont:
This completely breaks MakeMaker on both both Windows and VMS because it's depending on certain methods to be inherited from `MM_Any`, not `MM_Unix`. [I agree](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/91) our inheritance insane, but the solution is far more involved than this.

Leont:
Also, for the record, this is our inheritance tree:
~~~~ { .text }
               PACK### (created each call to ExtUtils::MakeMaker->new)
                    .                       |
                 (mixin)                    |
                    .                       |
        MY (created by ExtUtils::MY)        |
             |                              |
        ExtUtils::MY         MM (created by ExtUtils::MM)
                   |          |
                   ExtUtils::MM
                    |     |   |----------------+/\/\/\/\/\/\/\/\/\/\/\/\/+
                    |     |                    | Left path is chosen for |
     ExtUtils::Liblist   ExtUtils::MakeMaker   | VMS, Windows, AIX, VOS  |
     |        |                        |       | darwin, BeOS, Cygwin    |
     |  ExtUtils::Liblist::Kid     Exporter    | QNX, OS/2 and Netware   |
     |                                         | All others are a plain  |
     |                                         | Unix                    |
     |                                         |                         |
     |                                     ExtUtils::MM_{Current OS}     |
     |                                       |              |            |
     |                                       |            ExtUtils::MM_Unix
     |                                       |              |
     |                                       |     +--------+
     |                                       |     |
     |                                  ExtUtils::MM_Any
     |                                       |
     +---------------------------------------+
                    |
                File::Spec
~~~~

rurban:
I see. I need to add `use mro 'dfs'` then. Will do

Leont:
> I see. I need to add use mro 'dfs' then. Will do

Yeah, that sounds like a sensible solution to your problem (we can't really do that on our side because mro is a 5.10ism).

rurban:
Nope. mro is XS and EUMM needs to be run with miniperl.
Will rather fix #91 then. Sounds easier.

PR #287, opened 2017-02-14
Fix multiple inheritance for c3

rurban:
With multiple inheritance (@ISA>1), the c3 mro
refuses to load the most specific after the least specific base class.
The general rule to avoid inconsistency for @ISA is
from most specific to least specific.

cperl switched to c3 already (perl5 might do later evtl.)
and fails with "Inconsistent hierarchy during C3 merge".

(The other culprit is usually Exporter at the front of an @ISA btw).


rurban:
Oops, sorry. Didn't see that you already rejected it,

Issue #288, opened 2017-02-15
Security: parse_version eval


rurban:
get_version is dangerous code

Beware of the security issues for cpan testers or unchecked sources, as
the '$VERSION = ...'  line in a .pm file is simply evaluated, without
any sanity check.

Allowed can only a $ver_qr with `/^v? [\d_\.]+/` and `/qv\($ver_qr\)/`,
but for sure not any sub call or ; or eval

We cannot use Safe as it needs to run with miniperl, but at least a security sanity check is required.

Leont:
Yes, this is known. There is no way to do this safely and correctly (people do crazy stuff in their versions), one has to choose :-/

rurban:
The worst I've found with http://grep.cpan.me/?q=%5C%24VERSION+%3D+%5Cw
are 
```
Encode-InCharset-0.03/InCharset.pm:
our $VERSION = do { my @r = (q$Revision: 0.3 $ =~ /\d+/g); sprintf "%d."."%02d" x $#r, @r };

Compress/Raw/Bzip2.pm:
$VERSION = '0.0_01';
$XS_VERSION = $VERSION; 
$VERSION = eval $VERSION;
```

* The eval idiom can be detected in the XS _ case, and be left out. There's no need to eval a VERSION which was already defined.

* id expansion is not a thing anymore, but some still do it. $Revision: 0.3 parsing can be easily added.

* $VERSION = Other->version not needed for EUMM.

* our $VERSION = version->declare('v0.29.0');
needs to be changed to $VERSION = v0.29.0; which is not backcompat. qv() neither.
but these are not used in a VERSION_FROM file, or in packages >5.10.

All others should safely die.

karenetheridge:
We're working towards having Module::Metadata extract and evaluate the version in a Safe compartment, and it is the hope that EUMM's version extraction will be replaced with a call to Module::Metadata.

karenetheridge:
our $VERSION = '0.0_01';
    $VERSION = eval $VERSION;

This is an important construct and should not be removed.  The *single line* containing the $VERSION definition is detected and extracted by Module::Metadata, and should contain the underscore.  Performing an eval on a separate line ensures that $VERSION in running code is numeric and does not contain an underscore.

    our $VERSION = version->declare('v0.29.0');

I believe this was the recommended way to declare a v-string version, but others are more qualified to comment here.

[...]

PR #291, opened 2017-03-03
parse versions statically when possible

haarg:
This is incomplete, but is a start at addressing #288.

Parsing version declarations by evaling them is rather unsafe.  It would
be possible to improve the security of that by using a Safe container.
However, Safe.pm is an XS module and EUMM has to work without any XS.
Safe.pm should only be unavailable when compiling perl core though, and
the versions in perl core can all be parsed statically.

This patch takes the first step, by implementing static parsing of
versions that follow common patterns.  Later, a Safe container can be
added for the eval code path.  We may consider adding a warning for
the more exotic forms, and possibly rejecting them in the far future.

There are currently some inconsistencies.  Primarily, unquoted versions are extracted as strings rather than their parsed values.  So the following version declarations will be parsed differently:
```
$VERSION = 0.0;
$VERSION = 1.200_003;
$VERSION = v1.2;
$VERSION = 1.2.3;
```

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -2904,23 +2904,66 @@ sub parse_version {
     return $result;
 }
 
+my $v = qr{v?[0-9._]+};
+my $_quoted_version = qr{
+  \s*
+  (?:
+      (['"]?) ($v) \1
+    | qq? \s* (?:
+      | ([^\s\w]) ($v) \3
+      | \s ([\w]) ($v) \5
+      | \( ($v) \)
+      | \< ($v) \>
+      | \[ ($v) \]
+      | \{ ($v) \}
+    )
+  )
+  \s*
+}x;
+
 sub get_version {
     my ($self, $parsefile, $sigil, $name) = @_;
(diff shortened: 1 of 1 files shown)


karenetheridge:
Does it make sense to do this all in EUMM, as opposed to bundling a copy of
(a subset of) Module::Metadata to do this work?  Ideally we should just
have one code path that can do a robust job of version parsing, so
everything is consistent.

We're pretty close to being able to get the branch merged which uses a Safe
container, I think.

haarg:
I don't love the idea of adding another prereq to EUMM.

haarg:
@karenetheridge Maybe we could move this code (and any further version parsing code) into an ExtUtils::ParseVersion or something that could be packaged with EUMM and both EUMM and Module::Metadata could use it.  Then we can share the code without dealing with more bundling or prereqs in EUMM.

karenetheridge:
@haarg that would be eminently feasible, and make it a lot easier to wrangle future fixes to Module::Metadata as well.

I'll take some of my rambling thoughts to \#toolchain for now (and we have the hackathon coming up soon too to work things out)..

PR #293, opened 2017-04-15
Only add static libraries that are installed under auto/

Leont:
Some dists install their own static libraries to the archlib, which leads to build failures for `makeaperl`. This patch tries to eliminate that problem by only linking in static libraries in `$arch/auto/`.

### DIFF
--- a/t/03-xsstatic.t
+++ b/t/03-xsstatic.t
@@ -11,9 +11,6 @@ use MakeMaker::Test::Utils;
 use MakeMaker::Test::Setup::XS;
 use Test::More;
 
-plan skip_all => "Disabled as broken perl installs give false negative"
-  # if not static perl, and not author
-  unless !$Config{usedl} or $ENV{AUTHOR_TESTING};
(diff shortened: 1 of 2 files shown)


t/03-xsstatic.t
```diff
@@ -11,9 +11,6 @@ use MakeMaker::Test::Utils;
 use MakeMaker::Test::Setup::XS;
 use Test::More;
 
-plan skip_all => "Disabled as broken perl installs give false negative"
-  # if not static perl, and not author
-  unless !$Config{usedl} or $ENV{AUTHOR_TESTING};
```
mohawk2: I recall a great deal of discussion that ended with default-disabling the static-build tests with this. What's the rationale behind default-enabling?

Leont: The rationale is that this patch fixes said false negatives. I've tested this with Alien::unibilium and Alien::SVN.

mohawk2: In that case I'm in favour!

PR #294, opened 2017-05-09
Correct test description.

jkeenan:
At this point, 'testdir' is write-only (0200), not read-only.

### DIFF
--- a/t/eu_command.t
+++ b/t/eu_command.t
@@ -151,13 +151,6 @@ BEGIN {
         is( ((stat('testdir'))[2] & 07777) & 0700,
             0100, 'change a dir to execute-only' );
 
-        # change a dir to read-only
-        @ARGV = ( '0400', 'testdir' );
-        ExtUtils::Command::chmod();
-
-        is( ((stat('testdir'))[2] & 07777) & 0700,
-            0400, 'change a dir to read-only' );
-
         # change a dir to write-only
         @ARGV = ( '0200', 'testdir' );
         ExtUtils::Command::chmod();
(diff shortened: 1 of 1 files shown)


haarg:
I believe the correct thing to do is to reverse the read only and write only checks, so that the rm_rf is test description is correct.  That's what this was originally intended to check (fa708b9e2200b643549c9f302e77b99341fc743c), but it somehow got lost in the shuffle between EU::Command, EUMM, and blead.

jkeenan:
On 05/09/2017 03:35 AM, Graham Knop wrote:
> I believe the correct thing to do is to reverse the read only and write
> only checks, so that the rm_rf is test description is correct. That's
> what this was originally intended to check (fa708b9
> <https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/fa708b9e2200b643549c9f302e77b99341fc743c>),
> but it somehow got lost in the shuffle between EU::Command, EUMM, and blead.
>

Done.  Please re-review.

Thank you very much.
Jim Keenan

haarg:
Could you squash these into a single commit?  Other than that, looks good.

jkeenan:
On 05/09/2017 10:18 AM, Graham Knop wrote:
> Could you squash these into a single commit? Other than that, looks good.
>

Pushed.

jkeenan:
If this pull request is satisfactory, I would request a merge into master and timely release of a new version of ExtUtils-MakeMaker to CPAN.

I have to soon release a new version of File-Path.  It's doing fine on CPANtesters, but there would be one failure in ExtUtils-MakeMaker's t/eu_command.t in the absence of a new release of the latter.

Thank you very much.

Jim Keenan

PR #303, opened 2017-05-29
prune .a-finding in auto/share/*

mohawk2:
Attempt at fixing https://rt.cpan.org/Ticket/Display.html?id=121918


Leont:
I think it should prune `auto/share`, not just `auto/share/dist`. Otherwise module-sharedirs wouldn't be handled correctly.

mohawk2:
Agreed. Have also updated with @haarg 's shorter version of the match code.

PR #307, opened 2017-07-04
Filter out non-XS .a files

mohawk2:


mohawk2:
@Leont I believe this would complete what's needed to make static builds work when non-XS `.a` files get installed.

PR #311, opened 2017-12-12
Document when more attributes are available

Grinnz:


Grinnz:
I've found a few more keys that need such documentation... will add to the PR.

Grinnz:
Of note: the EUMM version did *not* strip the underscore at runtime until version 6.57_07, hence specifying it for versions before then and not after. I specified dev versions because those are sometimes used in perl releases, but I can change them to the first stable version where the feature was available if preferred.

Grinnz:
Also, I only added versions for features added after EUMM 6.17, which came with perl 5.6.2 and 5.8.1.

Leont:
Yeah this is useful to have!

mohawk2:
Great idea!

PR #312, opened 2017-12-31
Reinstate %Recognized_Att_Keys

cpansprout:


lib/ExtUtils/MakeMaker.pm
```diff
@@ -1783,6 +1783,11 @@ The following attributes may be specified as arguments to WriteMakefile()
 or as NAME=VALUE pairs on the command line. Attributes that became
 available with later versions of MakeMaker are indicated.
 
+A computer-readable list of recognized attributes is available as
```
mohawk2: "introspectable" instead?

Issue #119, opened 2014-08-04
Top-level README.pod installed as manpage


karenetheridge:
As seen in https://github.com/moose/MooseX-Role-WithOverloading/issues/3

haarg:
@mohawk2 That commit only seems to address manifying a README.pod.  It will still install the pod files themselves.

mohawk2:
If you'd like to reopen this for further investigation, I will further investigate. A test that fails would be super!

karenetheridge:
The issue is still open, or I would do so now :)

FGasper:
This is biting me with IO::Framed, IO::SigGuard, and Net::WebSocket.

FGasper:
PR submitted.

mohawk2:
@karenetheridge I reckon with #313 being merged and released, this can be closed?

PR #313, opened 2018-01-09
Teach libscan() to exclude README.pod.

FGasper:
This fixes issue #119.

### DIFF
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -2899,13 +2899,20 @@ Takes a path to a file or dir and returns an empty string if we don't
 want to include this file in the library.  Otherwise it returns the
 the $path unchanged.
 
-Mainly used to exclude version control administrative directories from
-installation.
+Mainly used to exclude version control administrative directories
+and base-level F<README.pod> from installation.
 
 =cut
 
 sub libscan {
     my($self,$path) = @_;
+
+    if ($path =~ m<^README\.pod$>i) {
+        my $ref = ref $self;
+        warn "WARNING: Older versions of $ref may errantly install $path as part of this distribution. It is recommended to avoid using this path in CPAN modules.\n";
+        return '';
+    }
+
     my($dirs,$file) = ($self->splitpath($path))[1,2];
     return '' if grep /^(?:RCS|CVS|SCCS|\.svn|_darcs)$/,
                      $self->splitdir($dirs), $file;
--- a/t/MM_Unix.t
+++ b/t/MM_Unix.t
@@ -12,7 +12,7 @@ BEGIN {
         plan skip_all => 'Non-Unix platform';
     }
     else {
-        plan tests => 110;
+        plan tests => 113;
     }
 }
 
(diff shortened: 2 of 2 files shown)


karenetheridge:
let's not be in a hurry to merge this. I'm not quite convinced yet it's the right action to take.

FGasper:
@karenetheridge Caution is certainly reasonable, but I’m also curious as to why this has lain unresolved for so long. Is there some workaround, or is just no one putting README.pod into their distros?

My own workaround, for IO::Framed and IO::SigGuard, was to hacky-mock libscan() to reject README.pod.

Without that, because my team converts all installed Perl modules to RPMs, we actually couldn’t install those two modules concurrently because of this bug.

TBH, I’m not wild about the fix I’ve submitted, but it at least constrains the scope to the very specific problem as reported and as appears to have affected several distributions.

FGasper:
@karenetheridge Is there something I can do to follow up on your concern?

karenetheridge:
> is just no one putting README.pod into their distros?

Definitely not; the problem does still exist in the wild. Many tickets have been filed though and some distributions have been cleaned up.

Is it possible for the RPM creation process to filter these files out?

FGasper:
@karenetheridge The RPM process can definitely filter out `/README.pod`, but that takes discovery of the issue each time, and each person who makes a CPAN RPM then will have to learn about this issue and its resolution. Meanwhile, each distro that wants to avoid it has to implement its own workaround. It would seem better for everyone if EU::MM were fixed.

I totally agree with erring on the side of caution for such a critical piece of the toolchain, but I’d also really like to see this fixed. :)

karenetheridge:
I mean, can't the RPM process itself always filter out README.pod?  What scripts are being used here? We can modify those.

FGasper:
I believe it’s a home-grown CPAN-to-RPM toolchain. (It’s not of my authorship, but I can ask.) Yeah, we can accommodate it, but someone on this end is going to ask, “well, why doesn’t EU::MM just fix its bug rather than having us accommodate it?”

Is it that the stakes for changing EU::MM are too high to warrant fixing this?

Grinnz:
I would say that 1. the stakes for changing anything in EUMM are extremely high, and 2. the payoff for changing things in EUMM usually does not apply to the majority of perl users who will still be using old versions of EUMM for years to come. That said I think this is worth cleaning up, as README.pod inclusion is a very common bug and there's really no way for users to know this is a problem until they're told about it.

toddr:
> I would say that 1. the stakes for changing anything in EUMM are extremely high, and 2. the payoff for changing things in EUMM usually does not apply to the majority of perl users who will still be using old versions of EUMM for years to come. 

I always have trouble with this argument. If you're saying nobody is going to use the new EU::MM for quite some time, you can't say the stakes are as high :)

My question would be what is the risk of applying the suggested patch? What could break?

Grinnz:
> I always have trouble with this argument. If you're saying nobody is going to use the new EU::MM for quite some time, you can't say the stakes are as high :)

I did not say nobody would use it; many people use new EU::MM versions as soon as they come out, as evidenced by everything that broke in 7.12.

bingos:
I am happy to merge this as a special case, as from #119 it seems to be affecting a lot of distributions

Leont:
> the stakes for changing anything in EUMM are extremely high

If someone genuinely has a `Foo::Readme` module and splits his documentation from his code, this would indeed break. I'm fine with a root `Readme.pod` being skipped, but I don't think it should be skipped elsewhere (`lib/` in particular), which I think this code would do.

> I always have trouble with this argument. If you're saying nobody is going to use the new EU::MM for quite some time, you can't say the stakes are as high :)

Yeah. The actual logic is the other way around: you should assume a lot of people won't upgrade. It means that CPAN authors should still not put a `README.pod` in the root directory. Being loud about having to skip this is probably a good idea.

sshaw:
I just finally used a modern version of ExtUtils::MakeMaker and saw this warning.  I wonder if this only became an problem because people started adding `README.pod`s for their code's GitHub page to their distro? At least that's when I started to notice it. 

Anyways, I have `README.*` in my `INSTALL.SKIP`. Seems that in this case the warning should not be generated?

PR #316, opened 2018-04-01
use tr instead of eval on $VERSION

karenetheridge:
This way, trailing zeroes are preserved in the version. This might be
important in some cases e.g. in metadata.


haarg:
Generally metadata should be extracted from the initial `$VERSION` definition, and would preserve the underscores and any trailing zeros.

But I do agree with making this change.  It preserves trailing zeros in error messages from `->VERSION` checks. `tr/_//d` is also faster, although that really doesn't matter for something like this.

karenetheridge:
One place I saw this ($ExtUtils::MakeMaker->VERSION) showing up is in META.json: e.g.

       "generated_by" : "ExtUtils::MakeMaker version 7.3, CPAN::Meta::Converter version 2.150010",

PR #317, opened 2018-04-22
Allow specifying extra inputs for PL_FILES scripts

ilmari:
The extra inputs are added as dependencies to the Makefile target, and
passed to the script after the target file name.


mohawk2:
Looks good to me.

Grinnz:
It should be indicated that this key does not take a hashref until a certain version of EUMM. Or perhaps it should be added as a new key.

Grinnz:
Either way anyone using this is going to need to provide backcompatibility code for older EUMM.

ilmari:
@Grinnz Or they can just specify the required EUMM version in `configure_requires`.

Issue #58, opened 2013-04-29
rewrite #!/usr/bin/env perl


miyagawa:
(Somewhat related to #37 and #38?)

Is there any reason that ExtUtils::MakeMaker fixin() rewrites `#!/usr/bin/perl` or `#!perl` but NOT `#!/usr/bin/env perl`?

1) There are many reasons `#!/usr/bin/env perl` do not work when installed. Keeping that line in the shebang sounds wrong to me.

2) Yet, having `#!/usr/bin/env perl` is useful for scripts that you want to run from a git repository etc. Look what I do with cpanm bin scripts so that you can run from a git repo by cp, but also installable to the right perl path, by rewriting the shebang in Makefile.PL: https://github.com/miyagawa/cpanminus/blob/devel/Makefile.PL#L9

There might be a valid use case for allowing `#!/usr/bin/env perl` as stated in #38 - so it might make sense to "allow" it by leaving as is if `@INC` is configured as relocatable, but by default, it makes more sense to rewrite it upon install, the same way it does with `#!perl`.

Grinnz:
There were some really good ideas near the end of this thread :+1: I just had another couple scripts installed with /usr/bin/env that then require manual intervention or gymnastics because they're not in my path.

neilb:
Just ended up here, because `#!/usr/bin/env perl` wasn't "corrected" on install.

Like the proposal of this being fixed by default (as I expect most people happy & not surprised by that), with an option to turn it off.

@mohawk2, @bingos: how's your supply of tuits? :-)

mohawk2:
I'll be pleased to write the probably small amount of code required, but I would like clear buy-in as to exactly what the rules are to be.

* Rewrite `#!perl` to `#!$PATH_TO_PERL`?
* Rewrite `#!/usr/bin/perl` to `#!$PATH_TO_PERL`?
* Rewrite `#!/usr/bin/env perl` to `#!$PATH_TO_PERL`?
* Have an option to rewrite any of the above to `#!/usr/bin/env perl`?
* Apply to all in `bin`? All in `scripts`?

Grinnz:
The first two can be simplified to encompass the current "rule" of rewriting anything that has the initial command ending in `perl`.

ap:
> Here's what I propose.
> 
> 1. Add a switch to turn fixin on or off (defaulting to on).
> 2. Change fixin so it will see #!/usr/bin/env perl as perl.
[...]

Issue #38, opened 2012-10-29
fixin for relocatable perls


schwern:
@shadowcat-mst [brought up](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/37#issuecomment-9523936) that fixin be changed to accommodate relocatable perls.

His suggestion that '#!/usr/bin/env perl' be allowed is less than ideal.  Programs installed with MakeMaker typically are associated with a module, possibly a particular version of a module.  There's a good chance that "whatever perl comes first in your PATH" isn't going to be the right one.  Alternatively, it requires careful management of what your current Perl is.

Matt, why don't you explain the problem some more please and let's see if we can figure out something?

Leont:
I almost thought I found a solution to this issue: hashbangs may contain a relative path instead of an absolute one. Sadly, this path is relative to the current working directory, because that ever makes sense -_-

dolmen:
@Leont Do not expect relative hashbangs to work on all platforms.

PR #322, opened 2018-06-28
Override shebang with PERL_MM_SHEBANG=relocatable

sjn:
- Relocatable shebangs are nice. Let's try to make them happen!
- This PR is related to tickets #38 and #58

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -1248,7 +1248,10 @@ sub _fixin_replace_shebang {
 
     # Now look (in reverse) for interpreter in absolute PATH (unless perl).
     my $interpreter;
-    if ( $cmd =~ m{^perl(?:\z|[^a-z])} ) {
+    if ( defined $ENV{PERL_MM_SHEBANG} && $ENV{PERL_MM_SHEBANG} eq "relocatable" ) {
+        $interpreter = "/usr/bin/env perl";
+    }
+    elsif ( $cmd =~ m{^perl(?:\z|[^a-z])} ) {
         if ( $Config{startperl} =~ m,^\#!.*/perl, ) {
             $interpreter = $Config{startperl};
             $interpreter =~ s,^\#!,,;


mohawk2:
Agreed, @sjn could you PR some docs?

haarg:
It's not clear to me why this is a separate environment variable rather than part of the normal EUMM options.

sjn:
@haarg: do you have a suggestion for a meaningful name? I was thinking of PERL_MM_OPT, PERL_MM_USE_DEFAULT when I picked this. Maybe "PERL_MM_FORCE_SHEBANG"?

@mohawk2, @Leont: Not entirely sure where documentation for this feature is naturally belongs... ExtUtils::MM_Unix's fixin() docs, or ExtUtils::MakeMaker's ENVIRONMENT section? Both?

haarg:
I think it should be just a parameter given to Makefile.PL.  Like `perl Makefile.PL SHEBANG=relocatable` or `export PERL_MM_OPT="SHEBANG=relocatable"`.  It shouldn't be a separate environment variable.

sjn:
@haarg: Ah, sorry, I misunderstood you. I assumed you didn't like the name. :smile:.

Personally I made this PR just to get the issue moving forward. It's been in limbo for FAR too long.
I chose an environment variable for this just because it was the smallest change possible to get something working.

I'll gladly try to throw in an attempt at this, but if I can have some assistance at writing tests, I'd be very happy. :grin:.

sjn:
@haarg: Ah, sorry, I misunderstood you. I assumed you didn't like the name. :smile:.

Personally I made this PR just to get the issue moving forward. It's been in limbo for FAR too long.
I chose an environment variable for this just because it was the smallest change possible to get something working.

I'll gladly try to throw in an attempt at using EUMM options, but if I can have some assistance at writing tests, I'd be very happy. :grin:.

mohawk2:
My feeling is that an environment variable is the wrong way to control this, but I'm not sure what's the right way. Perhaps a step towards figuring out the right way is to understand what problem we're really trying to solve.

It seems to me that installing a script with `env` in its shebang means it can fail if the wrong `perl` is first in its path. This sort of implies that maybe EUMM should instead forbid and override `env` shebangs, because that would guarantee runnable installed scripts. To try to bring this to being a single yes/no question: is there a situation where the `env` shebang is actively a good idea?

sjn:
Well, since EUMM does everything useful through the Makefiles it generates, I'd say that using environment variables is *well* within expectations for getting shit done. :smile: .

It's a bit of a kludge to make it work through the regular EUMM options (I'm not familiar with the code), but I'll take a stab at it when I find some tuits.

PR #324, opened 2018-07-17
Liblist::Kid tests, and on Unix handle dirs-with-spaces

mohawk2:
This re-implements testing this important module on Unix / OS/2, and uses `hash2files` to make the Win32 files, instead of awkwardly storing them in ` t/liblist/win32`.

Also implements the same use of `Text::ParseWords` on Unix as already on Win32 (i.e. not adding a dep) to handle any `-L` that uses quotes to protect spaces. On the output side, only quote-wraps any dirs that have spaces in.

### DIFF
--- a/lib/ExtUtils/Liblist/Kid.pm
+++ b/lib/ExtUtils/Liblist/Kid.pm
@@ -49,8 +49,10 @@ sub _unix_os2_ext {
     # $potential_libs
     # this is a rewrite of Andy Dougherty's extliblist in perl
 
+    require Text::ParseWords;
+
     my ( @searchpath );    # from "-L/path" entries in $potential_libs
-    my ( @libpath ) = split " ", $Config{'libpth'} || '';
+    my ( @libpath ) = Text::ParseWords::quotewords( '\s+', 0, $Config{'libpth'} || '' );
     my ( @ldloadlibs, @bsloadlibs, @extralibs, @ld_run_path, %ld_run_path_seen );
     my ( @libs,       %libs_seen );
     my ( $fullname,   @fullname );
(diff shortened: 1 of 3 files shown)


lib/ExtUtils/Liblist/Kid.pm
```diff
@@ -49,8 +49,10 @@ sub _unix_os2_ext {
     # $potential_libs
     # this is a rewrite of Andy Dougherty's extliblist in perl
 
+    require Text::ParseWords;
+
     my ( @searchpath );    # from "-L/path" entries in $potential_libs
-    my ( @libpath ) = split " ", $Config{'libpth'} || '';
```
Leont: Are you sure this is correct? In my understanding, $Config{libpth} is (currently) never quoted, and treating it as if it is will not result in a correct end result.

lib/ExtUtils/Liblist/Kid.pm
```diff
@@ -88,6 +90,7 @@ sub _unix_os2_ext {
                 $thislib = $self->catdir( $pwd, $thislib );
             }
             push( @searchpath, $thislib );
+            $thislib = qq{"$thislib"} if $thislib =~ / /; # protect spaces if there
```
Leont: I don't think this is correct quoting, it handles spaces but what about quotes or backslashes?

t/Liblist_Kid.t
```diff
@@ -58,6 +58,19 @@ sub conf_reset {
     delete $Config{$_} for keys %Config;
     %Config = %save_config;
     $Config{installarchlib} = 'lib';
+    # The following are all used and always are defined in the real world.
```
Leont: Shouldn't these just have been added to `@save_keys`?

lib/ExtUtils/Liblist/Kid.pm
mohawk2: This was intended as protection for the future. The only way it can go wrong currently is if there are accidental quote-characters in there, which will cause huge problems already.

lib/ExtUtils/Liblist/Kid.pm
mohawk2: I agree with your point. Are you aware of a converse to `Text::ParseWords` that reverse?

lib/ExtUtils/Liblist/Kid.pm
Leont: If the encoding of libpth ever changes (to support spaces in paths), then we can easily (perl) version check to handle it.

PR #330, opened 2018-11-13
Avoid mangling /bin non-perl shebangs on merged-/usr systems

ilmari:
If the shebang is absolute and exists in PATH, but was not the first
one found, leave it alone if it's actually the same file as first one.

This avoids packages built on merged-/usr systems with /usr/bin before
/bin in the path breaking when installed on systems without merged
/usr.  See e.g. https://bugs.debian.org/913637.

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -1256,8 +1256,8 @@ sub _fixin_replace_shebang {
     my ( $self, $file, $line ) = @_;
 
     # Now figure out the interpreter name.
-    my ( $cmd, $arg ) = split ' ', $line, 2;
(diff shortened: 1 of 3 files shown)


mohawk2:
It seems like a good idea to have tests for this?

Grinnz:
This seems awfully special-cased... That said I'm not sure the purpose of mangling non-perl shebangs to begin with...

ilmari:
@mohawk2 the whole PATH-related bit of shebang rewriting is currently completely untested, other than to check that it doesn't rewrite a non-existing non-perl interpreter.

ilmari:
@mohawk2 tests added

mohawk2:
Seen. They look great.

PR #332, opened 2019-02-16
Add.appveyor+fixes

rivy:
EU:MM has been broken on MSWin32 with early versions of perl (specifically Strawberry Perl < v5.14) for quite some time:

- for perl versions < 5.14, installing ExtUtils::MakeMaker >= 7.0 causes ... "The system cannot find the path specified.\ndmake.exe:  Error code 129, while making 'blib\lib\ExtUtils\.exists'"
- for perl versions = 5.12, installing ExtUtils::MakeMaker >= 7.0 causes ... "Use of inherited AUTOLOAD for non-method Win32::GetConsoleCP() is deprecated at C:/strawberry/perl/site/lib/ExtUtils/MakeMaker/Locale.pm line 50."
- for perl versions <= 5.10, installing ExtUtils::MakeMaker < 6.98 or >= 7.0 causes various odd failures (eg, "! Installing ExtUtils::MakeMaker failed. ... Retry with --force to force install it.").

For perl versions > 5.24, the default make type was changed to `gmake` and ExtUtils::MakeMaker works without issue.

I created an improved AppVeyor CI config (in the spirit of @haarg's Travis CI toolset) which I've been testing and using for a while and now I'm slowly releasing it into more public use. The supplied AppVeyor config (and related helper tools) is able to test all the way back to Perl v5.8.8.8 (the earliest Strawberry Perl available). Further information about the AppVeyor CI config is available at [rivy/CI.AppVeyor.helpers-perl](https://github.com/rivy/CI.AppVeyor.helpers-perl/blob/master/README.mkd).

I wanted to get EU:MM working back that far as well (see failures on report at [AppVeyor CI/add.appveyor](https://ci.appveyor.com/project/rivy/perl-extutils-makemaker/builds/18502150)). So, I created this PR which includes that AppVeyor CI (for testing) along with several annotated commits that fix all the MSWin32-related problems.

The PR compiles successfully all the way back to Perl v5.8.8.8 for MSWin32/AppVeyor and Perl v5.6.2 for Travis (see reports at [AppVeyor CI](https://ci.appveyor.com/project/rivy/perl-extutils-makemaker/builds/22425643) and [Travis CI](https://travis-ci.org/rivy/perl.ExtUtils-MakeMaker/builds/494315234)). It also includes support for code coverage (see reports at [CodeCov](https://codecov.io/gh/rivy/perl.ExtUtils-MakeMaker/branch/add.appveyor%2Bfixes) and [Coveralls](https://coveralls.io/github/rivy/perl.ExtUtils-MakeMaker)).

I know it's a relatively large set of changes, but I've the segregated the changes into fairly small, mentally digestible components.

I hope this is acceptable, and let me know if you'd like any changes.

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -1103,7 +1103,6 @@ Finds the executables PERL and FULLPERL
 
 sub find_perl {
     my($self, $ver, $names, $dirs, $trace) = @_;
-
     if ($trace >= 2){
         print "Looking for perl $ver by these names:
 @$names
@@ -2077,6 +2076,11 @@ sub init_PERL {
     # already escaped spaces.
     $self->{FULLPERL} =~ tr/"//d if $Is{VMS};
 
+    # `dmake` can fail for image (aka, executable) names which start with double-quotes
+    # * push quote inward by at least one character (or the drive prefix, if present)
+    # * including any initial directory separator preserves the `file_name_is_absolute` property
+    $self->{FULLPERL} =~ s/^"(\S(:\\|:)?)/$1"/ if $self->is_make_type('dmake');
+
     # Little hack to get around VMS's find_perl putting "MCR" in front
     # sometimes.
     $self->{ABSPERL} = $self->{PERL};
@@ -2098,6 +2102,11 @@ sub init_PERL {
     # Can't have an image name with quotes, and findperl will have
     # already escaped spaces.
     $self->{PERL} =~ tr/"//d if $Is{VMS};
+
+    # `dmake` can fail for image (aka, executable) names which start with double-quotes
+    # * push quote inward by at least one character (or the drive prefix, if present)
+    # * including any initial directory separator preserves the `file_name_is_absolute` property
+    $self->{PERL} =~ s/^"(\S(:\\|:)?)/$1"/ if $self->is_make_type('dmake');
 
     # Are we building the core?
     $self->{PERL_CORE} = $ENV{PERL_CORE} unless exists $self->{PERL_CORE};
--- a/lib/ExtUtils/MakeMaker/Locale.pm
+++ b/lib/ExtUtils/MakeMaker/Locale.pm
@@ -99,6 +99,13 @@ sub _init {
 	$ENCODING_LOCALE ||= $ENCODING_CONSOLE_IN;
     }
 
+    # Workaround of Encode < v2.71 for "cp65000" and "cp65001"
+    # The "cp65000" and "cp65001" aliases were added in [Encode v2.71](https://github.com/dankogai/p5-encode/commit/7874bd95aa10967a3b5dbae333d16bcd703ac6c6)
+    # via commit <https://github.com/dankogai/p5-encode/commit/84b9c1101d5251d37e226f80d1c6781718779047>.
+    # This will avoid test failures for Win32 machines using the UTF-7 or UTF-8 code pages.
+    $ENCODING_LOCALE = 'UTF-7' if lc($ENCODING_LOCALE) eq "cp65000";
+    $ENCODING_LOCALE = 'utf-8-strict' if lc($ENCODING_LOCALE) eq "cp65001";
+
     if ($^O eq "darwin") {
 	$ENCODING_LOCALE_FS ||= "UTF-8";
     }
--- a/lib/ExtUtils/MakeMaker/Locale.pm
+++ b/lib/ExtUtils/MakeMaker/Locale.pm
@@ -47,7 +47,10 @@ sub _init {
             unless (defined &GetInputCP) {
                 eval {
                     require Win32;
-                    eval { Win32::GetConsoleCP() };
+                    eval {
+                        local $SIG{__WARN__} = sub {} if ( "$]" < 5.014 ); # suppress deprecation warning for inherited AUTOLOAD of Win32::GetConsoleCP()
+                        Win32::GetConsoleCP();
+                    };
                     # manually "import" it since Win32->import refuses
                     *GetInputCP = sub { &Win32::GetConsoleCP } if defined &Win32::GetConsoleCP;
                     *GetOutputCP = sub { &Win32::GetConsoleOutputCP } if defined &Win32::GetConsoleOutputCP;
--- a/t/parse_version.t
+++ b/t/parse_version.t
@@ -110,7 +110,7 @@ our $VERSION = 2.34;
 END
 }
 
-if ( "$]" > 5.009 && "$]" < 5.012 ) {
+if ( "$]" < 5.012 ) {
   delete $versions{'$VERSION = -1.0'};
 }
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,173 @@
+# .appveyor.yml (for Perl distributions)
+
+# * (APIv1) for usage instructions, see <https://github.com/rivy/CI.AppVeyor.helpers-perl/blob/stable.APIv1/README.mkd>
+
+# ref: <https://www.appveyor.com/docs/appveyor-yml>[`@`](https://archive.is/OUJHS)
+# * "appveyor.yml" validation tool @ <https://ci.appveyor.com/tools/validate-yaml>
+
+version: "{build} ~ {branch}"
+
+branches:
+  except:
+    - gh-pages
+
+skip_tags: true     ## do not build on tags
+
+matrix:
+  allow_failures:
+    - AUTHOR_TESTING: 1
+    - TEST_SIGNATURE: 1
+    - CI_HELPER_BRANCH: canary
+    - Perl_VERSION: "5.8.8" ## allow perl v5.8.8 failures; (v5.8.8 is problematic for many projects)
+
+environment:
+  matrix:
+    - Perl_VERSION: "latest"
+      COVERAGE: "Codecov Coveralls" ## note: case sensitive!
+      AUTHOR_TESTING: 1
+      RELEASE_TESTING: 1
+    - Perl_VERSION: "5.28"
+    - Perl_VERSION: "5.26"
+    - Perl_VERSION: "5.24"
+    - Perl_VERSION: "5.22"
+    - Perl_VERSION: "5.20"
+    - Perl_VERSION: "5.16"
+    - Perl_VERSION: "5.14"
+    - Perl_VERSION: "5.12"
+    - Perl_VERSION: "5.10"
+    - Perl_VERSION: "5.8.9"
+    - Perl_VERSION: "5.8.8"
+    - Perl_VERSION: "latest"
+      CI_HELPER_BRANCH: "canary"
+      TEST_SIGNATURE: 1
+  global:
+    AUTOMATED_TESTING: 1
+    CI_CACHE_DIR: "C:\\cache"
+    #
+    # debug?
+    # CI_DEBUG: 1               ##  [ "" == false, <non-empty> == true ]; "true" enables detailed output for the helper scripts
+    #
+    ## configuration (CI/network-side settings and then ".appveyor_init.{BAT,PS1}" [if present], may override these values [by design])
+    ## ---
+    # * standard perl distribution test configuration signals
+    AUTHOR_TESTING: ""
+    RELEASE_TESTING: ""
+    # * .appveyor.yml configuration variables
+    CI_HELPER_API_VERSION: "1"  ## * API version (the expected/requested helper API version)
+    CI_HELPER_BRANCH: ""        ## "" => auto-set, valid alternates are [ "stable" (the default), "canary", BRANCH, or TAG ]
+    CI_HELPER_REPO: ""          ## "" => auto-set; allows easier use of alternate helper scripts (ie, alternate forks)
+    DEVEL_COVER_OPTIONS: ""     ## "" => auto-set, value determined by DIST_TOOLING; quoted whitespace (eg, " ") is an empty/neutral setting which also blocks auto-set
+    DIST_EXTRA_DEPS: ""         ## additional required/requested dependencies for build and/or testing
+    DIST_SUPPRESS_DEPS: ""      ## [ "" == false, <non-empty> == true ]; "true" can be useful for CORE modules, suppressing discovery and installation of dependencies, unless otherwise required (by COVERAGE or DIST_EXTRA_DEPS)
+    DIST_TOOLING: ""            ## [ "build", "make" ]; "" => auto-set based on existence of "Build.PL" and/or "Makefile.PL"
+    TEST_METHOD: ""             ## "" => auto-set based on DIST_TOOLING (`perl Build test`, `%make% test`, or `prove -bl`)
+    TEST_FILES: ""              ## "" => auto-set to "" for build/make distributions, otherwise "t" or "t xt" depending on AUTHOR_TESTING and/or RELEASE_TESTING and directory existence
+    ## ---
+
+cache:
+# note: unused, because, unfortunately, the available cache is too small to be helpful (unshared; 1GB across all projects)
+#   - '%CI_CACHE_DIR% -> .appveyor.yml'
+
+install:
+  - ps: ; write-host $("[{0:HH:mm:ss}].install" -f $($mark = get-date; $mark)) -f darkgray
+  # ensure CWD is project main directory
+  - cd "%APPVEYOR_BUILD_FOLDER%"
+  # perform any special preparation (optional; distribution specific)
+  # * note: optional external file(s), global scope => allows ".appveyor.yml" to remain more stable and similar [or the same] between different distributions [eg, for automated distribution of ".appveyor.yml" changes]
+  - if EXIST ".appveyor_init.BAT" ( call .appveyor_init.BAT )
+  - ps: if ( test-path ".appveyor_init.PS1" ) { . ".\.appveyor_init.PS1" }
+  # save current-point-in-time environment (for later logging of environment variable overrides)
+  - ps: $baseline_env = @(get-childitem env:)
+  # create a working area
+  - ps: if ( ! $env:CI_TEMP_DIR ) { $env:CI_TEMP_DIR = "${env:TEMP}\${env:APPVEYOR_JOB_ID}" ; mkdir -force $env:CI_TEMP_DIR | out-null }
+  # create cache area, if missing
+  - if NOT DEFINED CI_CACHE_DIR ( set "CI_CACHE_DIR=%CI_TEMP_DIR%\cache" )
+  - if NOT EXIST "%CI_CACHE_DIR%" ( mkdir "%CI_CACHE_DIR%" )
+  # finalize HELPER repository location
+  - if NOT DEFINED CI_HELPER_API_VERSION ( set "CI_HELPER_API_VERSION=1" )
+  - if NOT DEFINED CI_HELPER_REPO ( set "CI_HELPER_REPO=https://github.com/rivy/CI.AppVeyor.helpers-perl.git" )
+  - if NOT DEFINED CI_HELPER_BRANCH ( set "CI_HELPER_BRANCH=stable" )
+  - if /i "%CI_HELPER_BRANCH%"=="canary" ( set "CI_HELPER_BRANCH=canary.APIv%CI_HELPER_API_VERSION%" )
+  - if /i "%CI_HELPER_BRANCH%"=="stable" ( set "CI_HELPER_BRANCH=stable.APIv%CI_HELPER_API_VERSION%" )
+  # find / download helpers
+  - ps: ; write-host $("[{0:HH:mm:ss}].install (helpers)" -f $($mark_sub = get-date; $mark_sub)) -f darkgray
+  # * use "vendored" copy if present
+  - ps: if ( ! $env:CI_HELPERS -and (test-path ".appveyor_bin") ) { $env:CI_HELPERS=[IO.Path]::GetFullPath(".appveyor_bin") ; ${env:CI_HELPER_REPO} = ${env:APPVEYOR_REPO_NAME} ; ${env:CI_HELPER_BRANCH_DESC} = ${env:APPVEYOR_REPO_COMMIT} ; }
+  # * clone/download helpers from HELPER repo (via `git`) if needed
+  - ps: if ( ! $env:CI_HELPERS ) { $repo_path = $( mkdir "${env:CI_TEMP_DIR}\helpers" ).Fullname ; $git_cmd = "git clone ${env:CI_HELPER_REPO} `"${repo_path}`" -b ${env:CI_HELPER_BRANCH} 2>&1" ; write-host "[``${git_cmd}``]" ; & 'CMD' @( '/c', $git_cmd ) ; $err = $LASTEXITCODE ; $env:CI_HELPERS = "${repo_path}\.appveyor_bin" ; if ($err -ne 0) { exit $err } ; }
+  - ps: if ( ! $env:CI_HELPER_BRANCH_DESC ) { ${env:CI_HELPER_BRANCH_DESC} = $( pushd "${env:CI_HELPERS}" ; & 'git' @( 'describe', '--always' ) 2>&1 ; popd ) ; }
+  - ps: ; write-host $("[{0:HH:mm:ss}].install (helpers) ... (${env:CI_HELPER_REPO}@${env:CI_HELPER_BRANCH_DESC})")
+  # resolve requested coverage with needed configuration
+  - ps: . "${env:CI_HELPERS}\#install.determine-coverage.PS1"
+  ##
+  # highlight any overrides within baseline environment
+  - ps: $baseline_env | & "${env:CI_HELPERS}\log.env-overrides.PS1"
+  ##
+  # move "C:\mingw" to avoid cross library linking (a problem with older perl versions, breaking dll compilation with "/mingw/lib/dllcrt2.o:(.text+0xd1): undefined reference to `__dyn_tls_init_callback'")
+  # * only truly needed for modules containing XS compilation, but ok for all modules
+  - move c:\mingw c:\mingw.o >NUL
+  # force branch checkout (if knowable), then reset to the specific commit ## (needed for accurate code coverage info)
+  # * this allows later apps to see the branch name using standard `git branch` operations, yet always builds the correct specific commit
+  # * ref: <https://github.com/appveyor/ci/issues/1606>[`@`](https://archive.is/RVpnF)
+  - if DEFINED APPVEYOR_REPO_BRANCH if /i "%APPVEYOR_REPO_SCM%"=="git" ( git checkout "%APPVEYOR_REPO_BRANCH%" >NUL & git reset --hard "%APPVEYOR_REPO_COMMIT%" )
+  # install perl (strawberry variant) + version verification
+  - ps: ; write-host $("[{0:HH:mm:ss}].install (perl)" -f $($mark_sub = get-date; $mark_sub)) -f darkgray
+  - ps: . "${env:CI_HELPERS}\#install.install-perl.PS1"
+  - perl -V
+  - ps: ; write-host $("[{0:HH:mm:ss}].install (perl) ... done ({1:0.0}s)" -f $(get-date; $($(get-date) - $mark_sub).totalseconds)) -f gray
+  # determine build tooling
+  - ps: . "${env:CI_HELPERS}\#install.determine-tooling.PS1"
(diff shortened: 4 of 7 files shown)


lib/ExtUtils/MakeMaker/Locale.pm
```diff
@@ -47,7 +47,10 @@ sub _init {
             unless (defined &GetInputCP) {
                 eval {
                     require Win32;
-                    eval { Win32::GetConsoleCP() };
+                    eval {
+                        local $SIG{__WARN__} = sub {} if ( $] lt '5.014' ); # suppress deprecation warning for inherited AUTOLOAD of Win32::GetConsoleCP()
```
karenetheridge: This should be `if "$]" < '5.014'`

t/parse_version.t
```diff
@@ -110,7 +110,7 @@ our $VERSION = 2.34;
 END
 }
 
-if ( $] > 5.009 && $] < 5.012 ) {
+if ( $] lt '5.012' ) {
```
Grinnz: Similarly should remain a numerical comparison.

lib/ExtUtils/MakeMaker/Locale.pm
rivy: Changed to `$] < 5.014` (numeric, but more similar to the use throughout the rest of the module).

I'll force push the result once it passes CI.

t/parse_version.t
rivy: Fixed.
I'll force push it once CI completes successfully.

lib/ExtUtils/MakeMaker/Locale.pm
Leont: Should it? `$]` is a string value, not a numeric value…

rivy: I'm happy replace the comparison with whatever idiom is the best/most correct.

Looking at the repo, `$] < ...` is what is used throughout (except within my proposed changes):

```
lib/ExtUtils/MM_Any.pm:1093:        my $p2m = sprintf <<'CMD', $section, $] > 5.008 ? " -u" : "";
lib/ExtUtils/MM_Cygwin.pm:89:        if( $] >= 5.006002 ) {
lib/ExtUtils/MM_NW5.pm:195:                              $] =~ /(\d)\.(\d{3})(\d{2})/;
lib/ExtUtils/MM_Unix.pm:2918:    if ( $pod_encoding and !( $] < 5.008 or !$Config{useperlio} ) ) {
lib/ExtUtils/MM_Unix.pm:3197:    if ($] >= 5.008) {
lib/ExtUtils/MM_VMS.pm:208:                push(@cand,$name) if $name =~ /^[\w\-\$]+$/;
lib/ExtUtils/MM_VMS.pm:220:        if ($name =~ /^[\w\-\$]+$/) {
lib/ExtUtils/MakeMaker.pm:401:    push @Get_from_Config, qw( vendorarchexp vendorlibexp ) if $] >= 5.006;
lib/ExtUtils/MakeMaker.pm:537:        !$self->{MIN_PERL_VERSION} or $self->{MIN_PERL_VERSION} <= $]
lib/ExtUtils/MakeMaker.pm:547:            die sprintf <<"END", $self->{MIN_PERL_VERSION}, $];
lib/ExtUtils/MakeMaker.pm:555:                $self->{MIN_PERL_VERSION}, $];
lib/ExtUtils/MakeMaker.pm:593:          $pr_version = $];
lib/ExtUtils/MakeMaker.pm:1269:        utf8::encode $to_write if !$CAN_DECODE && $] > 5.008;
lib/ExtUtils/MakeMaker/Locale.pm:51:                        local $SIG{__WARN__} = sub {} if ( $] lt '5.014' ); # suppress deprecation warning for inherited AUTOLOAD of Win32::GetConsoleCP()
lib/ExtUtils/MakeMaker/version.pm:38:        if ($] >= 5.009000) {
lib/ExtUtils/MakeMaker/version/vpp.pm:937:	if ( $] ge 5.008_001 ) {
lib/ExtUtils/MakeMaker/version/vpp.pm:941:	elsif ( $] ge 5.006_000 ) {
lib/ExtUtils/MakeMaker/version/vpp.pm:976:    if ( exists $INC{"$class.pm"} and not %{"$class\::"} and $] >= 5.008) {
lib/ExtUtils/MakeMaker/version/vpp.pm:985:	local $^W if $] <= 5.008;
lib/ExtUtils/MakeMaker/version/vpp.pm:992:	    my $msg =  $] < 5.006
lib/ExtUtils/Mksymlists.pm:151:    unless($] >= 5.016) {
```

My preference would be use the same comparision and then change them all in another commit.

Thoughts?

Grinnz: @Leont it is a string value, but the correct way to compare it is numeric (or creating version.pm objects).

Leont: What would go wrong the other way? This is rather counter-intuitive to me.

rivy:
Version comparison changes completed.

lib/ExtUtils/MakeMaker/Locale.pm
Grinnz: With carefully constructed strings nothing should go wrong either way, but it is in principle a numeric comparison - it should still work if your version string is '005.008' for example

karenetheridge: `$]` should be converted to a string first, and then compared numerically. Zefram wrote about this a few years back documenting the various issues that can arise when compared otherwise.

I prepared a patch to fix this everywhere, in #333.

t/build_man.t
```diff
@@ -44,10 +44,15 @@ ok((my $stdout = tie *STDOUT, 'TieOut'), 'tie stdout');
 
 {
     local $Config{installman3dir} = File::Spec->catdir(qw(t lib));
-    my $mm = WriteMakefile(
-        NAME            => 'Big::Dummy',
-        VERSION_FROM    => 'lib/Big/Dummy.pm',
-    );
+    my $mm;
```
karenetheridge: Why is this commit needed? Where is README.pod coming from?

rivy:
Per @karenetheridge 's notes, version comparisons are converted to `"$]" < ...` form.

A reference to Zefram discussing at least one of the issues is at <https://rt.perl.org/Public/Bug/Display.html?id=127909#txn-1396084>.

I've rebased on top of #333 ([wip@rivy/perl-MakeMaker](https://github.com/rivy/perl.ExtUtils-MakeMaker/tree/wip)). I'll re-push after #333 is merged.

t/build_man.t
rivy: It's needed to suppress the given warning that occurs in all MSWin32 tests ... see <https://ci.appveyor.com/project/rivy/perl-extutils-makemaker/builds/20544004>.

As a test, it a harmless change that cleans up the output and suppresses the spurious WARNINGS. I don't see the WARNING or file locally when tested. I can gate is specifically to AppVeyor if you see some problem with it.

rivy: I re-tested it without the suppression, and it occurs locally as well.
There is no *README.pod* file to interfere, just the plain README file. It just spits out the spurious warnings. I'm don't know why it occurs on MSWin32, but not *nix.

Leont:
> Per @karenetheridge 's notes, version comparisons are converted to `"$]" < ...` form.

You converted it to use numeric operators, but didn't add the explicit stringification. The stringification is the essential part for that approach to work because…

> A reference to Zefram discussing at least one of the issues is at https://rt.perl.org/Public/Bug/Display.html?id=127909#txn-1396084.

That issue involves using floating points for versions.

The string values of `$]` have always worked sensibly (or at least they do on 5.6.2, the oldest perl on my computer): `$] ge '5.008'` will DWIM on any perl version. It's semantics are clear and correct.

On the other hand `"$]" >= '5.008'` will give the same result because both sides go through the same (#72210 arised because the floats were generated through different processes). This should give the same result as using a stringwise operator, but performs more operations (first a conversion to string, then a conversion to number). It doesn't seem like the logical solution to me.

t/testrules.yml
```diff
@@ -0,0 +1,11 @@
+---
+# TAP::Harness test rules
+# "t\02-xsdynamic.t" (and possibly "t\03-xsstatic.t") should *not* be run in parallel
```
Leont: This should not be racy. If it is that's a bug we should fix and not something we should work around.

lib/ExtUtils/MakeMaker/Locale.pm
```diff
@@ -47,7 +47,10 @@ sub _init {
             unless (defined &GetInputCP) {
                 eval {
                     require Win32;
-                    eval { Win32::GetConsoleCP() };
+                    eval {
+                        local $SIG{__WARN__} = sub {} if ( $] < 5.014 ); # suppress deprecation warning for inherited AUTOLOAD of Win32::GetConsoleCP()
```
Leont: `local $^W if …`?

t/fixin.t
```diff
@@ -127,6 +127,7 @@ END
 SKIP: {
     eval { chmod(0755, "usrbin/interp") }
         or skip "no chmod", 8;
+    skip "unable to chmod into executable", 8 if not -x "usrbin/interp"; # Win32 files are executable by virtue of file extension (not chmod permissions)
```
Leont: This doesn't make sense to me

Grinnz:
> The string values of `$]` have always worked sensibly (or at least they do on 5.6.2, the oldest perl on my computer): `$] ge '5.008'` will DWIM on any perl version. Its semantics are clear and correct.

I still have to disagree. Here is another example why this is fundamentally not a string comparison. `$] ge '5.008000'` - is this true on 5.008? It depends on whether $] stringifies to `5.008` or `5.008000`. Whichever is the case, `$] gt '5.008'` will have the opposite problem.

t/build_man.t
Grinnz: I think this PR should not concern itself with these warnings, and if there is a bug with them on Win32 that should be fixed separately.

rivy:
Ok, the various version comparison recipes all seem to work and pass CI, even my original version. The current PR is in the form `$] > ...` (bare `$]` with a numeric comparison to a number). @karenetheridge just posted #333 which converted the entire module to `"$]" > ...` (stringified `$]` with numeric comparison to a number). I have a version of the PR based on that form as well ([wip@rivy/perl.ExtUtils-MakeMaker](https://github.com/rivy/perl.ExtUtils-MakeMaker/tree/wip)).

I'm happy to change the comparison to whichever is the "best practice". I just need a consensus.

t/fixin.t
rivy: Sorry, but it doesn't work on Windows.
```
C:> echo > file

C:> perl -E "eval { chmod(0755, 'file') } or print 'chmod failed:'; if (-x 'file') { print 'executable:'} else {print 'NOT executable:'}; print qq/done/"
NOT executable:done
C:>ver

Microsoft Windows [Version 10.0.17763.292]

C:>perl -v

This is perl 5, version 28, subversion 1 (v5.28.1) built for MSWin32-x64-multi-thread

Copyright 1987-2018, Larry Wall

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at http://www.perl.org/, the Perl Home Page.
```

Leont:
> I still have to disagree. Here is another example why this is fundamentally not a string comparison. `$] ge '5.008000'` - is this true on 5.008? It depends on whether $] stringifies to `5.008` or `5.008000`.

All ranged comparisons work correctly if you do the sensible and lazy thing and write no more decimals than necessary (but if you really want to know, in 5.10.0 we switched to always use 6 decimals in `$]`, but 5.8.0 is even less supported than 5.6.2 so the point is moot anyway).

> Whichever is the case, `$] gt '5.008'` will have the opposite problem.

If would be false for 5.8.0 in both cases, and true for 5.8.1 ;-)

Grinnz:
No, '5.008000' gt '5.008' is true.

Leont:
> So given what you said, the 'sensible' approach will fail for gt on 5.10.0 or higher.

Right, though I still can't imagine anyone ever wanting to do that for at least two reasons: the only sensible operations on it are greater-or-equal-than or less-than; and you can just look up valid $] values if you really have to)

If there is any implementation where "5.008" numifies to a subtly different float than "5.008000" does (no idea if that exists but I'm pretty sure it'd be legal), using numeric operators this way could give "interesting" results. The logic works on the assumption that both sides will numify in the same way, 

But `$^V gt v5.10.0` will always (on 5.6+) do the right thing. We're at the point where `$^V` and vstrings suddenly look sane again.

lib/ExtUtils/MakeMaker/Locale.pm
rivy: Sure, I'm happy to rewrite it into that idiom. I see that that formulation is already is use elsewhere in the module (ie, *lib/ExtUtils/MakeMaker/version/vpp.pm*).

lib/ExtUtils/MakeMaker/Locale.pm
rivy: @karenetheridge , What your timeline for merging #333? It causes a (minor) conflict with this PR, so I rebased on top of it. I'll push the updated branch (which uses `"$]" > ...`) when #333 is in master.

ribasushi:
> We're at the point where `$^V` and vstrings suddenly look sane again.

`$^V` has other potential side effects on-use, let's not confuse this issue further. The proposed solution in #333 is the correct ( albeit unfortunate ) one.

lib/ExtUtils/MakeMaker/Locale.pm
rivy: Nope, the suppression is requiring the full $`SIG{__WARN__}...`.
I'm just going to leave it in full verbosity (and function).

t/testrules.yml
rivy: It is racy on Windows.

If you can point to a problem area to search, I'll spend some time looking into it. But it's intermittent (as these things are) which increases the difficulty of tracking it down.
Plus, it may be an artifact of the test harness and Windows, completely unrelated to the module. I don't want to go down a rabbit hole for something that hasn't occurred in actual use of the module or has an actual repeatable failing test. So, to me, serializing the tests seems to be a small price to pay to keep the purposeful tests from interfering with each other.

I can put in a FIXME comment and we can open as issue about it if you'd prefer.

t/build_man.t
rivy: Ok, so the "README.pod" file is left over from the creation on [line 41](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/332/files#diff-4473c5dc5a51f147c095c02ce3b60062L41).

There is no associated test, so I likely assumed at the time that it was needed for something and didn't touch it. 

Placing a following `unlink $README or die ...` does fix the problem, and, much more directly, just removing [lines 40-41](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/332/files#diff-4473c5dc5a51f147c095c02ce3b60062L40-41) completely fixes it as well.

I don't know why it's not a WARNING on *nix, just on Windows. And I don't know really why it's there to begin with, so I'm hesitant to just remove it. Although the tests run fine with it completely removed (for both Windows and *nix).

But, I do think that it needs to be fixed, either with the change I have here, the `unlink ...`, or complete removal of those two lines, as the WARNINGS are a useless distraction while testing in Windows.

rivy:
It's been a couple of days without further comments, so, to keep this from getting stale, I'm going to update the PR to include version comparisons in the `"$]" > ...` idiom and remove the extraneous lines which create the "README.pod" file (along with the now redundant work-around code).

All tests are passing.

lib/ExtUtils/MM_NW5.pm
```diff
@@ -192,7 +192,7 @@ MAKE_FRAG
     }
     # Reconstruct the X.Y.Z version.
     my $version = join '.', map { sprintf "%d", $_ }
-                              $] =~ /(\d)\.(\d{3})(\d{2})/;
+                              "$]" =~ /(\d)\.(\d{3})(\d{2})/;
```
Grinnz: This stringification isn't actually necessary, since it already is used only as a string here.

Grinnz: I see this is from #333, that has been merged to master so you should rebase to current master.

rivy: Done.

t/testrules.yml
Leont: Is it racy with MSVC or with GCC? Do you have example reports of failures

That test is easily the slowest in the test suite. It's responsible for 30% of the total test time on my machine, which means that on a 3-4 core system this would double the total testing time.

If windows is racy, it probably shouldn't be running parallel. I don't really like making that a problem for all other platforms.

rivy: It's Strawberry Perl, so it uses `gcc`.

I ran through my AppVeyor builds and can't find a report with the failure, but it was intermittent.

Maybe the best thing would be to drop the serialization of the tests and wait to see if/when it to reappears and address it at that time.

Leont: Yeah, I suspect that's best

rivy: Ok, I removed that commit. I'll hold it in a feature branch for future consideration if it comes back up again. Thanks for the input!

t/fixin.t
rivy: It looks like @bingos just added a commit skipping this test on MSWin32 in a slightly different manner (see <https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/85c6b588f29f52a1f3474dc5d52b7653ccacecb9>). So, I'll remove the commit with this change as well.

rivy:
I believe all voiced concerns have been addressed.

I'm force pushing an updated set of commits which pass all tests on both AppVeyor and Travis CIs.

(@bingos?) Is this now mergeable?

t/build_man.t
```diff
@@ -37,8 +37,6 @@ END {
 
 ok( chdir 'Big-Dummy', "chdir'd to Big-Dummy" ) ||
   diag("chdir failed: $!");
-my $README = 'README.pod';
-{ open my $fh, '>', $README or die "$README: $!"; }
```
haarg: The warnings this causes are not unique to Windows.  The tests later in the file are meant to validate that no extra man page files get generated, including making sure it doesn't generate them for a top level README.pod file.  Possibly the warnings should be suppressed rather than removing this.

rivy: Hmm, that's what I proposed initially with https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/1ebf6e6293cb842ee651d597ba82e621b3f36098 but @karenetheridge and @Grinnz objected in the conversation above... Is https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/1ebf6e6293cb842ee651d597ba82e621b3f36098 how you think it should be resolved? Or do you have another suggestion?

Grinnz: My suggestion was not to touch this at all in this PR since it's unrelated.

rivy: I see this PR as "add MSWin32 testing" and "fix the problems in that testing environment". This cleans up the testing, so, IMO, it's related.

Since the change is in a single commit, it could be be easily moved to another PR. But, for the same reason, including it here is also easy and fixes the distracting warnings.

Is the point that you want to debate whether this should be done or how it should be done in another thread?

mohawk2: > Is the point that you want to debate whether this should be done or how it should be done in another thread?

I think given the discussion above, that is a good idea. If it takes >1 PR to get Windows in the needed place, so be it.

rivy: * commit removed

rivy:
(@bingos ?) mergeable now?

lib/ExtUtils/MM_Unix.pm
```diff
@@ -2077,6 +2076,11 @@ sub init_PERL {
     # already escaped spaces.
     $self->{FULLPERL} =~ tr/"//d if $Is{VMS};
 
+    # `dmake` can fail for image (aka, executable) names which start with double-quotes
+    # * push quote inward by at least one character (or the drive prefix, if present)
+    # * including any initial directory separator preserves the `file_name_is_absolute` property
+    $self->{FULLPERL} =~ s/^"(\S(:\\|:)?)/$1"/ if $self->is_make_type('dmake');
```
mohawk2: Is there a reason not to use `Text::ParseWords` to unwrap, then some more conceptually correct thing, rather than just an RE?

lib/ExtUtils/MM_Unix.pm
```diff
@@ -2099,6 +2103,11 @@ sub init_PERL {
     # already escaped spaces.
     $self->{PERL} =~ tr/"//d if $Is{VMS};
 
+    # `dmake` can fail for image (aka, executable) names which start with double-quotes
+    # * push quote inward by at least one character (or the drive prefix, if present)
+    # * including any initial directory separator preserves the `file_name_is_absolute` property
+    $self->{PERL} =~ s/^"(\S(:\\|:)?)/$1"/ if $self->is_make_type('dmake');
```
mohawk2: Same point here

lib/ExtUtils/MM_Unix.pm
rivy: I hadn't considered that option.

I see your point that my solution here is just a mechanical work-around for the `dmake` bug. But, to me, conceptually, the FULLPERL string is correct, it's just that `dmake` can't handle an initial double-quote. So, a simple mechanical fix seems reasonable to me.

And, just quickly mentally spit-balling a variant solution using `Text::ParseWords`, I don't think that it would be an improvement. To my mind, splitting the string and reconstructing it into a `dmake`-palatable version (ie, sub-quoting pieces of the string) would be more error-prone.

I'm happy to be persuaded otherwise if you have a better construction in mind that you could share.

lib/ExtUtils/MM_Unix.pm
rivy: Same response here... 😄

lib/ExtUtils/MM_Unix.pm
mohawk2: I think that a parsewords, then a `File::Spec::splitdir`, then quoting the directory parts and filename using `MM->quote_literal`, would be the most reliable. REs have a way of going wrong.

rivy: Hmm...

``` perl
$self->{FULLPERL} =~ s/^"(\S(:\\|:)?)/$1"/ if $self->is_make_type('dmake');
```

would be replaced with 

``` perl
$self->{FULLPERL} = do {
    use File::Spec q//; 
    use Text::ParseWords q//;
    my $p = $self->{FULLPERL}; 
    $p =~ s/\\/\\\\/gmsx;
    my @w = Text::ParseWords::shellwords($p);
    join(q/ /, File::Spec->catpath(map {MM->quote_literal($_)} File::Spec->splitpath(shift @w)), map {MM->quote_literal($_)} @w);
    } if $self->is_make_type('dmake');
```

or, if we can assume that `$self->{FULLPERL}` is a always a single "shellword", 

``` perl
 $self->{FULLPERL} = do {
    use File::Spec q//; 
    use Text::ParseWords q//;
    my $p = $self->{FULLPERL}; 
    $p =~ s/\\/\\\\/gmsx;
    File::Spec->catpath(map {MM->quote_literal($_)} File::Spec->splitpath((Text::ParseWords::shellwords($p))[0]));
    } if $self->is_make_type('dmake');
```

I can do something like that, but I'm hesitant. Do you think that's a better version?

Grinnz: I have no experience to speak to what's more appropriate, but if you add use statements please do so at the top of the file or use require to lazy-load, it's misleading to put use statements in smaller scopes.

mohawk2: A runtime `require` would also achieve the same as the `use...()`, as well as the lazy aspect mentioned above.

mohawk2: Also I'm not sure of the purpose of the `\` related RE?

rivy: This was just a first pass to show the steps that I thought were going to be needed.

@grinnz , I'll ultimately push the modules up in scope or use `require ...`.

@mohawk2 , paths in Win32 use backslashes which unfortunately is used as an escape character within the `Text::ParseWords` functions. I don't see a way to turn that off, so I pre-escaped the backslashes.

Am I getting that you both believe this is a better way to go?

rivy: So, after testing and iterating on the code, it looks like this construction is what is required to replace the original RE (at both locations, with FULLPERL => PERL in the second location)...

``` perl
$self->{FULLPERL} = do {
    require Text::ParseWords;
    my $p = $self->{FULLPERL};
    $p =~ s/\\/\\\\/gmsx;
    my @w = Text::ParseWords::shellwords($p);
    if (!@w) { $p =~ s/^"//msx; @w = Text::ParseWords::shellwords($p); }
    join(q/ /, File::Spec->catpath(map {$self->quote_literal($_)} File::Spec->splitpath(shift @w)), map {$self->quote_literal($_)} @w);
    } if $self->is_make_type('dmake');
```

`$p =~ s/\\/\\\\/gmsx;` is required because `shellwords()` otherwise removes the backslashes from the path (commonly used for windows paths). And the ` if (!@w) { $p =~ s/^"//msx; @w = Text::ParseWords::shellwords($p); }` construction is required because some supplied FULLPERL/PERL strings have unbalanced double-quotes, which causes `shellwords()` to return an empty list. And the entire returned `shellwords()` array must be considered as perl flags may be attached to the string as a suffix.

I don't know of an available alternative to `shellwords()` which ignores backslash escapes and is forgiving of unbalanced double-quotes. If such a function exists, the code could be boiled down to a more palatable ...

``` perl
$self->{FULLPERL} = do {
    require ...;
    my @w = ALTERNATE_SHELLWORDS($self->{FULLPERL});
    join(q/ /, File::Spec->catpath(map {$self->quote_literal($_)} File::Spec->splitpath(shift @w)), map {$self->quote_literal($_)} @w);
    } if $self->is_make_type('dmake');
```

But, otherwise, I'm inclined to stick with the current RE. It's succinct and works in all tested `dmake` make-style perls.

I could limit it to just the MSWin32 `dmake` make-style perls, which are all tested via AppVeyor. That could limit any breakage on other perl variants (if any other `dmake` make-style variants actually exist), eg ...

``` perl
$self->{FULLPERL} =~ s/^"(\S(:\\|:)?)/$1"/ if ($self->is_make_type('dmake') && ($^O eq 'MSWin32'));
```

mohawk2: As the misguided person who made EUMM work with gmake on Windows, and with spaces in paths, I am somewhat familiar with its quirks. I'd forgotten that ParseWords is effectively recreating the Bourne (etc) shell, rather than being per-platform. That being so, the `\` stuff is quite reasonable. I would encourage you to consider making a Win32 equivalent of the full function (see `Win32::ParseWords`) to be as bulletproof as possible, rather than doing "one bit at a time", which is an easy rabbithole to get sucked into.

Anyway, the ultimate goal is to make something that actually works - being minimally ugly is above all a means towards that, rather than an end to itself. Keep up the good work.

rivy: Thanks for the feedback and conversation.

I'm going to leave this as is and resolve it, if that's ok with you.

Or, I can add the OS restriction (`$^O eq 'MSWin32'`) if you feel that narrowing the scope adds value. The OS restricted version still fixes the `dmake` bug and passes all tests.

I'll add creating a more portable `...::ParseWords` to my "ToDO list". Or, I might create a patch for `Text::ParseWords`, adding optional arguments and enabling more flexible use.

mohawk2: If the limitation being dealt with is in dmake, keep it specific to that. There's no reason for it to apply to Win32 with gmake, I gather. By the way, a future version (not to feature-creep this version) will want to also test against gmake. That only takes an edit of `Config_heavy.pl`, from memory, but let's first get something that operates!

rivy: @mohawk2 , FYI, the more recent versions of Strawberry Perl are using `gmake`. So, the current AppVeyor testing does test against both `dmake` and `gmake`.

rivy:
Mergeable now?

mohawk2:
Looks like it to me. Let's fix any issues that come up as a separate matter.

rivy:
It's been another couple of weeks, and I see that @bingos is adding commits to "master" in what looks like preparation for another release.

Is there any blocker to including this PR?

PR #338, opened 2019-07-07
Guess man section from man*dir

afresh1:
In OpenBSD we put perl module man pages into the "3p" section, but MM only knows to tell `pod2man` about section 1 or 3, not the actual section where they live.  This PR will attempt to detect the section based on the "install${INSTALLDIRS}man[13]dir", similar to what is done to detect the "man?ext" in perl's Configure script.

I'm not sure what limitations there are with this being a core module that is included with perl, but am open to suggestions.

### DIFF
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -1088,9 +1088,9 @@ manifypods : pure_all config $dependencies
 END
 
     my @man_cmds;
-    foreach my $section (qw(1 3)) {
-        my $pods = $self->{"MAN${section}PODS"};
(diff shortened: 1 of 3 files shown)


Leont:
What's the different between this and  `man1ext` / `man3ext`?

afresh1:
This is what gets passed to pod2man as the `--section` and what is embedded in the content of the page.  While often the same as `man?ext` I have been told that solaris puts things in a `man3` directory but uses `3perl` as the `man3ext`.  Unfortunately I don't have any other specific examples, but I believe using the `man?ext` would cause problems there.

PR #339, opened 2019-08-22
Fix static linking on macOS

jpc:
Perl searches for extensions to link statically by looking for `boot_*` symbols
inside the extension `.a` archives. This does not work on macOS since symbols
are prefixed with a `_` so a `_boot_*` pattern is needed. I believe the proposed pattern
should work on all platforms.


jpc:
Hi, this extension is shipped with (and used by) upstream Perl. Should I open a pull request against upstream Perl as well or will this propagate automatically?

Grinnz:
This is the correct location. Updates to this distribution will be pulled into core.

Leont:
I think cygwin would also need the underscore :-)

jpc:
Thanks. :)

PR #351, opened 2020-03-29
MM_Win32.pm: make can_dep_space check for short path support

richardleach:
MM_Win32.pm's **can_dep_space()** always assumes that short path name support is present. But this is increasingly not the case. (It is disabled by default on non-C: drives on modern Windows releases.)

This commit makes can_dep_space() test whether short path name support is present.

Doing so fixes test failures (when short path name support is not present) in 02-dynamic.t. These failures come from MM_Win32.pm's **quote_dep()** being unable to handle the "type map" argument in the absence of short path support.

"type map" comes from MakeMaker::Test::Setup::XS, which will actually remove the space if the make flavour cannot handle spaces - but this happening depends upon can_dep_space() being accurate:
```
my $typemap = 'type map';
my $MM = MM->new({NAME=>'name', NORECURS=>1});
$typemap =~ s/ //g unless $MM->can_dep_space;
```

### DIFF
--- a/lib/ExtUtils/MM_Win32.pm
+++ b/lib/ExtUtils/MM_Win32.pm
@@ -395,8 +395,15 @@ sub perl_script {
 }
 
 sub can_dep_space {
-    my $self = shift;
-    1; # with Win32::GetShortPathName
+    my ($self) = @_;
+    return 0 unless $self->can_load_xs;
+    require Win32;
+    require File::Spec;
+    my ($vol, $dir) = File::Spec->splitpath($INC{'ExtUtils/MakeMaker.pm'});
+    # can_dep_space via GetShortPathName, if short paths are supported
+    my $canary = Win32::GetShortPathName(File::Spec->catpath($vol, $dir, 'MakeMaker.pm'));
+    (undef, undef, my $file) = File::Spec->splitpath($canary);
+    return (length $file > 11) ? 0 : 1;
 }
 
 =item quote_dep


lib/ExtUtils/MM_Win32.pm
```diff
@@ -395,8 +395,11 @@ sub perl_script {
 }
 
 sub can_dep_space {
-    my $self = shift;
-    1; # with Win32::GetShortPathName
+    # can_dep_space via GetShortPathName, if short paths are supported
```
mohawk2, line 398: This needs a `require Win32;` for robustness.

richardleach: I have not gotten the knack of squashing commits yet...

haarg:
Is loading Win32 in this case safe?  One of the other places that uses a Win32 is guarded by a `can_load_xs` check.

richardleach:
> Is loading Win32 in this case safe? One of the other places that uses a Win32 is guarded by a `can_load_xs` check.

`quote_dep()` doesn't have that guard and that doesn't seem to be a problem. (If code has called `can_dep_space()`, it seems likely that it will currently call `quote_dep()` shortly thereafter.) `_normalize_path_name()` similarly also doesn't have that guard.

Moreover, the `can_load_xs()` guard in `arch_check()` was added for the benefit of miniperl during [building perl on Windows](https://rt.cpan.org/Public/Bug/Display.html?id=46667), as builds were otherwise broken.

I can build perl using _nmake_ and _gmake_ on Windows just fine without that guard being in place. (Don't have _dmake_ to hand.) So, adding that guard doesn't _seem_ necessary.

lib/ExtUtils/MM_Win32.pm
mohawk2: Can you make the indentation be exactly the same as the next line? I think you have a tab character.

mohawk2:
There is a whitespace issue (sigh) which I have commented on. @richardleach could you please also add the `can_load_xs` guard? This is the kind of thing I would prefer us not to have to go back and fix in a bit.

lib/ExtUtils/MM_Win32.pm
```diff
@@ -395,8 +395,12 @@ sub perl_script {
 }
 
 sub can_dep_space {
-    my $self = shift;
-    1; # with Win32::GetShortPathName
+	require Win32;
+    # can_dep_space via GetShortPathName, if short paths are supported
+    Win32::CreateFile("8dot3test_canary_can_dep_space $$");
```
mohawk2, line 400: Sorry for not raising this before. The current directory could be read-only at build time. Creating a file in the current directory does not seem to me correct. Please could you replace this by specifically checking the current filesystem for the shortname capability, rather than this approach?

richardleach: AFAIK there is no reliable way to specifically check a file system for the shortname capability without testing the actual behaviour.

Users with elevated privileges can create a handle to (e.g.) \\.\c​: and pass that handle to 
DeviceIoControl() in order to retrieve FILE_FS_PERSISTENT_VOLUME_INFORMATION, from which support or lack thereof can be determined. Users without those privileges - typical non-administrator users - cannot do that.

Would a viable approach be to try to read the shortname of a file that (1) doesn't fit the 8.3 naming convention and (2) comes with EUMM, perhaps _MakeMaker.pm_? If so, happy to amend this PR, but I'd like advice on what relative path construction to pass to Win32::GetShortPathName.

mohawk2: Yes, I think we can guarantee `MakeMaker.pm` would be there, and it's a clever approach. Use `File::Spec` to get the relative filename. Please also address the other issues raised.

richardleach: How does this look?

```
sub can_dep_space {
    my ($self) = @_;
    return 0 unless $self->can_load_xs;
    require Win32;
    require File::Spec;
    my ($vol, $dir, $file);
    ($vol, $dir, undef) = File::Spec->splitpath( $INC{'ExtUtils/MakeMaker.pm'} );
    # can_dep_space via GetShortPathName, if short paths are supported
    my $canary = Win32::GetShortPathName( File::Spec->catpath($vol, $dir, 'MakeMaker.pm') );
    (undef,undef,$file) = File::Spec->splitpath(  $canary );
    return ( length $file > 11 ) ? 0 : 1;
}
```

mohawk2: This looks fine. I'd nitpick the `my` could be done better (and I've also made the whitespace consistent):

```
sub can_dep_space {
    my ($self) = @_;
    return 0 unless $self->can_load_xs;
    require Win32;
    require File::Spec;
    my ($vol, $dir) = File::Spec->splitpath($INC{'ExtUtils/MakeMaker.pm'});
    # can_dep_space via GetShortPathName, if short paths are supported
    my $canary = Win32::GetShortPathName(File::Spec->catpath($vol, $dir, 'MakeMaker.pm'));
    (undef, undef, my $file) = File::Spec->splitpath($canary);
    return (length $file > 11) ? 0 : 1;
}
```

richardleach: Thanks. I've updated the PR to your version of _can_dep_space_.

lib/ExtUtils/MM_Win32.pm
richardleach: I've updated the PR. Please could you check - and let me know if you'd like me to squash the commits.

mohawk2: This now looks fine, but yes - please squash the commits. Else the merger can do a squash, but it's better practice by far to have one commit be a small, meaningful change, not a partial tweak of a previous commit.

mohawk2 (approved):
Apart from squashing together (which could be done on merge), this now seems fine.

lib/ExtUtils/MM_Win32.pm
richardleach: Now squashed.

PR #353, opened 2020-05-27
Always link to $Config{libs} on cygwin

Leont:
This is intended to fix Perl/perl5#17805

WIP because it needs testing. @pjacklam would you be able to do that?

### DIFF
--- a/lib/ExtUtils/MM_Cygwin.pm
+++ b/lib/ExtUtils/MM_Cygwin.pm
@@ -100,6 +100,16 @@ sub init_linker {
     $self->{EXPORT_LIST}  ||= '';
 }
 
+sub init_others {
+    my $self = shift;
+
+    $self->SUPER::init_others;
+
+    $self->{LDLOADLIBS} ||= $Config{perllibs};
+
+    return;
+}
+
 =item maybe_command
 
 Determine whether a file is native to Cygwin by checking whether it


tonycoz:
The cygwin failures look like they're because the system perl includes -lnsl in libs (and several others):

```
$ perl -V:libs
libs='-lpthread -lnsl -lgdbm -ldb -ldl -lcrypt -lgdbm_compat';
$ uname -a
CYGWIN_NT-10.0 phobos 3.1.4(0.340/5/3) 2020-02-19 08:49 x86_64 Cygwin
```

but a fresh build doesn't include them:

```
$ ./miniperl -Ilib -V:libs
libs='-lpthread -ldl -lquadmath';
```

(miniperl because of the issue this PR is fixing.)

bingos:
Yeah, that's on me, I'll fix the cygwin actions to pull in nsl

Leont:
Also, I was in doubt on wether to use libs or perllibs. I suspect the latter is more appropriate, but both are not very well defined.

karenetheridge:
related: https://rt.cpan.org/Ticket/Display.html?id=133808

Leont:
> Yeah, that's on me, I'll fix the cygwin actions to pull in nsl

That means that everyone else will have to do that as well, which appears to be a problem.

pjacklam:
The most recent versions of the perl Cygwin package have all had this -lnsl issue. Because of this, I have not been able to build any Perl module that uses XS code. However, if I build perl on Cygwin from source, I don't run into this issue. I am really at the edge of my knowledge her, but could this be a Cygwin problem rather than a perl problem? During Cygwin installation, the nsl package is not a prerequisite for the perl package, so I have no nsl library in my /lib directory on Cygwin.

Leont:
> The most recent versions of the perl Cygwin package have all had this -lnsl issue. Because of this, I have not been able to build any Perl module that uses XS code. However, if I build perl on Cygwin from source, I don't run into this issue. I am really at the edge of my knowledge her, but could this be a Cygwin problem rather than a perl problem?

if libnsl is present during perl's build, it will be included in the list of wanted libraries. So if perl was build with libnsl laying around (as cygwin apparently does), it needs to be around when extensions are build (which is rather unfortunate).

I think we need a different fix for this problem.

PR #356, opened 2020-07-21
Test oneliner without double escaping dollar sign

### LINKED ISSUE #355: oneliner helper does not allow variables
When reading the unit test for the `sub oneliner` from `t/oneliner.t`
```
try_oneliner(q{$PATH = 'foo'; print $PATH},[], q{foo},   'dollar signs' );
```

You have the feeling that oneliner authorizes the usage of variables, all the most the test is passing.
But in truth it's using references:
```
# $(ABSPERLRUN)  -e '$$PATH = '\''foo'\''; print $$PATH' --
```

I wonder if we should teach `sub oneliner` to `allow_variables` when calling `quote_literal`, or detect incorrect usage of variable...

For example you can use `%h` but then it's going to fail when using `$h` later to access a value... 

Detail:
====
`my %h = (1..4);` is a valid oneliner
```
$mm->oneliner( q[my %h = (1..4);]);
# $(ABSPERLRUN)  -e 'my %h = (1..4); print $$h[1]' --
```

but `my %h = (1..4); print $h[1]` is invalid
```
$mm->oneliner( q[my %h = (1..4);]);
# $(ABSPERLRUN)  -e 'my %h = (1..4); print $$h[1]' --
```

```perl
#!perl

use Test::More;

require ExtUtils::MM;

my $mm = MM->new(
	{ NAME => q[Fake], }
);

my $cmd;

# valid - using ref
my $cmd = $mm->oneliner( q[$PATH = 'foo'; print $PATH]);
note $cmd;
# $(ABSPERLRUN)  -e '$$PATH = '\''foo'\''; print $$PATH' --

# invalid
my $cmd = $mm->oneliner( q[my $PATH = 'foo'; print $PATH]);
note $cmd;
# $(ABSPERLRUN)  -e 'my $$PATH = '\''foo'\''; print $$PATH' --

```

atoomic:
Fix #355

Double escaping '$' for Makefile usage is fine, but
we should not use such a syntax while testing and trying
to run the oneliner output.

This change is 'unescaping' the '$$' to '$' so we can perform
some extra checks while using oneliner.

Otherwise as shown in this example a simple 'my $foo' test will fail.


Leont:
LGTM.

Technically it's wrong if the path to perl contains two dollar signs, moving it one-line up would fix that unlikely edge-case.

atoomic:
good point... was not considering it being an option, but you are right and I moved it above via db193818

mohawk2:
Generally I favour this change.

Unfortunately, the CI fail reveals a real problem: dmake has to double up its braces. If you want to unquote them, you'll need to reverse that too.

I feel like `try_oneliner` is trying to be too clever by trying to de-quote the make quoting, and it's getting it wrong. Surely the right thing to do is create small makefiles that have these recipes, and just use make to run them?

Leont:
> I feel like try_oneliner is trying to be too clever by trying to de-quote the make quoting, and it's getting it wrong. Surely the right thing to do is create small makefiles that have these recipes, and just use make to run them?

And apparently VMS also does its own unique thing. Writing out a Makefile seems like the way to go.

atoomic:
nice catch and this comes from the additional test using a hash I've added :-) 
let me see if unescaping those braces fixes it
```
syntax error at -e line 1, near "{{"
Execution of -e aborted due to compilation errors.

#   Failed test '%h and $h'
#   at t/oneliner.t line 42.
#          got: ''
#     expected: '2'
# oneliner:
# "C:\strawberry\perl\bin\perl.exe"  -e "my %h = (1, 2); print $h{{1}}" --
# Looks like you failed 1 test of 17.
[21:25:20] t/oneliner.t .............. 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/17 subtests 
```

atoomic:
going to work on the `Makefile` solution

Leont:
> going to work on the `Makefile` solution

On VMS, the Makefile needs to be called `Descrip.MMS` AFAIK. Possibly other operating systems have similar quirks.

atoomic:
I see `makefile_name` from t/lib/MakeMaker/Test/Utils.pm

craigberry:
> On Jul 22, 2020, at 11:13 AM, Nicolas R. <notifications@github.com> wrote:
> 
> 
> I see `makefile_name` from t/lib/MakeMaker/Test/Utils.pm

Right, use that one in the tests and in MakeMaker itself use $self->{MAKEFILE}.

> On Wed, Jul 22, 2020 at 10:11 AM Leon Timmermans <notifications@github.com>
> wrote:
> 
> > going to work on the Makefile solution
> >
> > On VMS, the Makefile needs to be called Descrip.MMS AFAIK. Possibly other
> > operating systems have similar quirks.
> 


________________________________________
Craig A. Berry

"... getting out of a sonnet is much more
 difficult than getting in."
                 Brad Leithauser

t/oneliner.t
```diff
@@ -7,55 +7,117 @@ BEGIN {
...
+
+    my $ok = $TB->is_eq( $output, $expect, $name )
+        || $TB->diag("$Makefile:\n$content");
+
+    return $ok;
+}
+
+sub Makefile_template {
+    my ($RUN) = @_;
+    my $NOECHO = '@';
```
Leont, line 107: This will not be correct on all platforms

atoomic: Indeed you are right. This is kind a of a chicken egg issue: producing Makefile to test code which generate Makefile.. 
What would be your preference? 
- do we have some helpers we could use for this?
- skipping some distro for this test?
- going back to the unescape solution?
- other?

thanks for your valuable input

mohawk2: Skipping and unescaping are not viable approaches. Probably the way to go here is to be making a sandbox dir, then minimal `Makefile.PL`. Look in the XS tests for functions to make this easy.

PR #358, opened 2020-07-28
use $^V instead of PERL_VERSION once possible (5.10)

toddr:
Due to issues with XS code assuming that the Perl major version is 5,
$CONFIG{PERL_VERSION} may be innacurate in a future version of Perl.
Instead, simply parse $^V and use that to produce the "ARCHITECTURE NAME"

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -3236,10 +3236,14 @@ PPD_PERLVERS
     }
 
     my $archname = $Config{archname};
-    if ("$]" >= 5.008) {
-        # archname did not change from 5.6 to 5.8, but those versions may
-        # not be not binary compatible so now we append the part of the
-        # version that changes when binary compatibility may change
+
+    # archname did not change from 5.6 to 5.8, but those versions may
+    # not be not binary compatible so now we append the part of the
+    # version that changes when binary compatibility may change
+    if( "$]" >= 5.010) {
+        $archname .= "-$^V->{version}->[0].$^V->{version}->[1]"; # v5.32.5 => 5.32
+    }
+    elsif ("$]" >= 5.008) {
         $archname .= "-$Config{PERL_REVISION}.$Config{PERL_VERSION}";
     }
     push @ppd_chunks, sprintf <<'PPD_OUT', $archname;
--- a/t/basic.t
+++ b/t/basic.t
@@ -132,8 +132,11 @@ like( $ppd_html, qr{^\s*<REQUIRE NAME="strict::" />}m,  '  <REQUIRE>' );
 unlike( $ppd_html, qr{^\s*<REQUIRE NAME="warnings::" />}m,  'no <REQUIRE> for build_require' );
 
 my $archname = $Config{archname};
-if( "$]" >= 5.008 ) {
-    # XXX This is a copy of the internal logic, so it's not a great test
+# XXX This is a copy of the internal logic, so it's not a great test
+if( "$]" >= 5.010) {
+    $archname .= "-$^V->{version}->[0].$^V->{version}->[1]";
+}
+elsif( "$]" >= 5.008 ) {
     $archname .= "-$Config{PERL_REVISION}.$Config{PERL_VERSION}";
 }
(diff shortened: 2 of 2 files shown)


karenetheridge:
`$^V` is a version object. We can use its interfaces to get the values of each segment, rather than using the stringification overload and then picking it apart again with a regex.

toddr:
> `$^V` is a version object. We can use its interfaces to get the values of each segment, rather than using the stringification overload and then picking it apart again with a regex.

Good idea. I don't see a method but the internals seem to be well documented, so dirty it is!

Leont:
> Due to issues with XS code assuming that the Perl major version is 5,
$CONFIG{PERL_VERSION} may be innacurate in a future version of Perl.

I get the point with doing that with the PerlAPI define, but why would we do the same with the %Config entry?

mohawk2:
The CI needs to be passing.

haarg:
This seems like something that would be better as a new value provided by Config for the "ABI compatible" version.

Poking the internals of version objects is definitely wrong. Unfortunately it doesn't provide access to the parts of the version, so using split would seem the proper solution if this approach was used.

Grinnz:
`$v->{version}[#]` is unfortunately the documented interface for version objects, despite the Internals label. It's safe to say that interface is now required to maintain backcompat for the forseeable eternity.

toddr:
> The CI needs to be passing.

The CI got confused because I submitted 2 back to back. a re-submit of the same code will fix it. back to the discussion first though.

toddr:
> I get the point with doing that with the perlapi define, but why would we do the same with the %Config entry?

When you look at the code in blead, it's clear that the macro and the Config variable are supposed to be the same. I've been having the following conversation with @atoomic:

Q: We're going to lie about the macro `PERL_VERSION` because we know that 99% of XS which uses it is wrong. Should we be doing it for `%Config` though without similar evidence?
R: I think it would be a mistake to have the values for `%Config` and `#define` macros differ. That'll make things even more confusing.

A quick search of CPAN shows that maybe 1/3 of CPAN would benefit from a lie on this in Config, 1/3 would prefer the truth, and another 1/3 are probably already fundamentally broken even now.

https://grep.metacpan.org/search?size=20&q=Config%7B%27PERL_VERSION%27%7D&qd=&qft=
https://grep.metacpan.org/search?size=20&q=Config%7BPERL_VERSION%7D&qd=&qft=

Luckily there's < 20 distros that even need this and a surprising amount seem to be copying around the code extracted from `EU::MM`. So I thought a good first pass would be to just fix EU::MM and see what that looks like.

toddr:
> Poking the internals of version objects is definitely wrong. Unfortunately it doesn't provide access to the parts of the version, so using split would seem the proper solution if this approach was used.

That is my instinct too, though the documentation seems to encourage doing so, so I would argue this is no longer internals: https://metacpan.org/pod/version::Internals#Version-Object-Internals

karenetheridge:
> We're going to lie about the macro `PERL_VERSION`

Can you clarify what this means?

toddr:
> > We're going to lie about the macro `PERL_VERSION`
> 
> Can you clarify what this means?

https://github.com/Perl/perl5/wiki/Perl-7,-XS-and-Devel::PPPort#patch-perl-7

toddr:
> > The CI needs to be passing.
> 
> The CI got confused because I submitted 2 back to back. a re-submit of the same code will fix it. back to the discussion first though.

I re-submitted and 5.12 is failing in appveyor. This really looks to be a failure in appveyor and not the commit since it's not even getting to the tests when the run fails.

Leont:
> This seems like something that would be better as a new value provided by Config for the "ABI compatible" version.

Yeah, a $Config{abi_version} or some such would be the helpful thing here

ribasushi:
> @bingos merged commit 6e58499 into Perl-Toolchain-Gang:master

This is not acceptable. A decent portion of #toolchain ( @Leont, @karenetheridge and @haarg ) all had reservations about this PR. An obligatory reminder that EU::MM is a Perl5 project, and thus not subject to the Perl7 "quality process".

@bingos please:
1. Revert/reopen for proper discussion to take place
2. Explain what steps are you taking to not have a repeat of this situation

PR #364, opened 2020-08-30
work around EU::CBuilder not working well in parallel

haarg:
ExtUtils::CBuilder->have_compiler properly uses unique file names for
some of its steps, but not all of them.  Change to a temporary directory
before using it to force its files to be unique.

This allows the two xs test scripts to run in parallel without issue, so
we can remove the test rules forcing them to run sequentially.


mohawk2:
It does look like the CI likes this (though AppVeyor is terribly slow).

PR #366, opened 2020-09-02
Make windows specific code for NoXS compile under strict

### LINKED ISSUE #365: t/lib/MakeMaker/Test/NoXS.pm does not compile on windows
As a result of the combination of 81d3c0c41d8005ee5ece0c01e0202c14f31c7605 and 67a6b5820e2f434db5ae52e3b06c0b522d454de3, this file no longer compiles on windows. I'll submit a PR with a fix in a few minutes.

```
C:\perl\cpan\ExtUtils-MakeMaker\t>perl -Ilib -MMakeMaker::Test::NoXS -e1
Can't use string ("Win32::GetCwd") as a symbol ref while "strict refs" in use at lib/MakeMaker/Test/NoXS.pm line 17.
Compilation failed in require.
BEGIN failed--compilation aborted.
```

toddr:
Fixes #365

### DIFF
--- a/t/lib/MakeMaker/Test/NoXS.pm
+++ b/t/lib/MakeMaker/Test/NoXS.pm
@@ -13,12 +13,18 @@ require XSLoader;
 delete $DynaLoader::{boot_DynaLoader};
 
 if ($^O eq 'MSWin32') {
+    # Load Win32. Then clear the stash of all other entries but GetCwd and SetChildShowWindow
+    # SetChildShowWindow and GetCwd are provided by perl core in modern perls, so we
+    # can use them in miniperl. on older perls, we can load them from Win32 so the
+    # test can proceed as normal.
+
     require Win32;
-    my $GetCwd = *{'Win32::GetCwd'}{CODE};
-    my $SetChildShowWindow = *{'Win32::SetChildShowWindow'}{CODE};
-    %{*main::Win32::{HASH}} = ();
-    *{'Win32::GetCwd'} = $GetCwd;
-    *{'Win32::SetChildShowWindow'} = $SetChildShowWindow;
+
+    foreach my $slot (keys %Win32::) {
+        next if $slot eq 'GetCwd';
+        next if $slot eq 'SetChildShowWindow';
+        delete $Win32::{$slot};
+    }
 }
 
 # This isn't 100%.  Things like Win32.pm will crap out rather than


Grinnz:
Looks correct (and much cleaner) to me

t/lib/MakeMaker/Test/NoXS.pm
```diff
@@ -13,12 +13,13 @@ require XSLoader;
...
 
 if ($^O eq 'MSWin32') {
+    # Load Win32. Backup GetCwd and SetChildShowWindow. Then clear the stash of all other entries.
     require Win32;
-    my $GetCwd = *{'Win32::GetCwd'}{CODE};
-    my $SetChildShowWindow = *{'Win32::SetChildShowWindow'}{CODE};
-    %{*main::Win32::{HASH}} = ();
-    *{'Win32::GetCwd'} = $GetCwd;
-    *{'Win32::SetChildShowWindow'} = $SetChildShowWindow;
+    my $GetCwd = Win32->can('GetCwd');
```
Leont, line 18: How about `\&Win32::GetCwd`?

toddr: Is this a stylistic comment or is there a difference?

Grinnz: It's a direct reference to the function so closer to what the original code did; there's no reason to care about inheritance here and the function names are static so no need for `->can` really.

t/lib/MakeMaker/Test/NoXS.pm
```diff
@@ -13,12 +13,13 @@ require XSLoader;
...
+    my $SetChildShowWindow = Win32->can('SetChildShowWindow');
+    %Win32:: = ();
+    *Win32::GetCwd = $GetCwd;
```
haarg, line 21: This doesn't work. It binds early, so the glob it's assigning to gets erased by the stash being cleared.

wolfsage:
Test

wchristian (changes requested):

t/lib/MakeMaker/Test/NoXS.pm
haarg: `*Win32::GetCwd` binds to the glob at compile time. Clearing the stash removes that glob from the namespace, but because it's already bound, the assignment stores the sub in a glob that isn't in the namespace anymore.

haarg:
```
delete @Win32::{ grep { $_ ne "GetCwd" && $_ ne "SetChildShowWindow" } keys %Win32:: };
```

t/lib/MakeMaker/Test/NoXS.pm
toddr: I have removed the use of can.

t/lib/MakeMaker/Test/NoXS.pm
toddr: Fair enough. Let's just not delete them to begin with as a solution?

toddr:
> delete @Win32::{ grep { $_ ne "GetCwd" && $_ ne "SetChildShowWindow" } keys %Win32:: };

Yep. saw this late. I decided the for loop would be more clear to people?

wchristian (approved):

wchristian:
Fwiw, i didn't mean to start a review. I am not sure how to use that interface. So i now clicked around on a few things and hope that made things better. :)

haarg:
Technically this looks fine.  It might be worth explaining why the deletion is being done. Possibly something like:
```
# SetChildShowWindow and GetCwd are provided by perl core in modern perls, so we
# can use them in miniperl. on older perls, we can load them from Win32 so the
# test can proceed as normal.
```

toddr:
> # SetChildShowWindow and GetCwd are provided by perl core in modern perls, so we
> # can use them in miniperl. on older perls, we can load them from Win32 so the
> # test can proceed as normal.

Part of my problem here was divining what @bulk88 's original intent was here. Is this for sure what's being done?

haarg:
Yes.

The test is to verify that EUMM works properly in miniperl, which is before any XS modules are available. SetChildShowWindow and GetCwd are provided directly by perl on modern perls, not by the Win32 module. So they are allowed to be used by EUMM under miniperl. However, older perls don't provide them. Loading the Win32 module, then removing all of the other symbols simulates the situation on newer perls, which allows the miniperl test to continue normally. The miniperl behavior is only needed at perl build time, so simulating a modern perl is acceptable for testing. Current EUMM won't ever be used for building older perls.

toddr:
comment updated.

PR #368, opened 2020-09-21
Make Mkbootstrap.pm strict.

toddr:
By explicitly referring to the DynaLoader package when referring to subs
or vars in that space, we can make Mkbootstrap.pm strict. This was already
being done in this package so we're just doing it in more places now.

### DIFF
--- a/lib/ExtUtils/Mkbootstrap.pm
+++ b/lib/ExtUtils/Mkbootstrap.pm
@@ -1,7 +1,6 @@
 package ExtUtils::Mkbootstrap;
 
-# There's just too much Dynaloader incest here to turn on strict vars.
-use strict 'refs';
+use strict;
 use warnings;
 
 our $VERSION = '7.47_11';
@@ -34,6 +33,7 @@ sub Mkbootstrap {
     if (-f "${baseext}_BS"){
 	$_ = "${baseext}_BS";
 	package DynaLoader; # execute code as if in DynaLoader
+	no strict 'vars';
 	local($osname, $dlsrc) = (); # avoid warnings
 	($osname, $dlsrc) = @Config::Config{qw(osname dlsrc)};
 	$bscode = "";
@@ -44,6 +44,7 @@ sub Mkbootstrap {
 
     if ($Config{'dlsrc'} =~ /^dl_dld/){
 	package DynaLoader;
+	no strict 'vars';
 	push(@dl_resolve_using, dl_findfile('-lc'));
     }
 


lib/ExtUtils/Mkbootstrap.pm
```diff
@@ -34,17 +33,15 @@ sub Mkbootstrap {
     if (-f "${baseext}_BS"){
 	$_ = "${baseext}_BS";
 	package DynaLoader; # execute code as if in DynaLoader
-	local($osname, $dlsrc) = (); # avoid warnings
-	($osname, $dlsrc) = @Config::Config{qw(osname dlsrc)};
-	$bscode = "";
+	($DynaLoader::osname, $DynaLoader::dlsrc) = @Config::Config{qw(osname dlsrc)};
```
Leont, line 36: Any particular reason for getting rid of the `local`?

Leont:
None of this is helpful unless the `_BS ` files it loads also becomes strict

lib/ExtUtils/Mkbootstrap.pm
toddr: I can put it back in but I'm unconvinced it actually works.

toddr:
> This doesn't solve the problem unless the `_BS ` files it loads also becomes strict, though its use is fairly rare nowadays

I agree. In point of fact, outside of blead there are only 3 BS files on CPAN at all. I'm unclear if the feature is redundant and mostly unnecessary these days.

This is the entire list of files from CPAN
```
D/DB_File/DB_File_BS
E/Embperl/Embperl_BS
H/HTML-Embperl/Embperl_BS
p/perl/cpan/DB_File/DB_File_BS
p/perl/ext/XS-APItest/APItest_BS
```

lib/ExtUtils/Mkbootstrap.pm
toddr: Actually putting the local back is a problem since DynaLoader never declares the variables to begin with. IMO the local is unnecessary in the scope we're talking about here.

craigberry:
Please keep in mind that Mkbootstrap mostly runs under miniperl during the core build and thus there is no DynaLoader.  The shenanigans with the DynaLoader namespace appear to be simulations of bits of DynaLoader in its absence. The patch may be ok but development on this module needs to be tested against the core build, not just against a checkout of MakeMaker.

toddr:
> this module needs to be tested against the core build, not just against a checkout of MakeMaker

I've been testing on blead already. I synced this here to get it into EUMM first.

toddr:
I can push a merge for this to blead first if you'd feel more comfortable with that @craigberry ?

craigberry:
I'm not concerned about where it gets pushed first -- just offering a reminder about things that have bitten folks in the past.

lib/ExtUtils/Mkbootstrap.pm
haarg: Why would that be a problem?

toddr:
After sleeping on it, I've decided to go with a more simple approach. It still smells to me that this code isn't really doing what we think it's doing, but that wasn't the problem I came here to fix. There are literally 5 modules in the world that use it so let's assume they work for now.

I am going to instead do `no strict 'vars';` in the 2 places that need it and move on.

PR #369, opened 2020-09-24
Skip the failing dynamic vs static tests on Win32 in blead

toddr:
Work around for https://github.com/Perl/perl5/issues/17601

See also https://rt.cpan.org/Ticket/Display.html?id=115321

This mirror's @tonycoz 's merge to blead found here https://github.com/Perl/perl5/pull/18157

### DIFF
--- a/t/lib/MakeMaker/Test/Setup/XS.pm
+++ b/t/lib/MakeMaker/Test/Setup/XS.pm
@@ -401,11 +401,17 @@ sub list_dynamic {
         $^O !~ m!^(VMS|aix)$! ? ([ 'subdirscomplex', '', '' ]) : (),
     ) : (), # DynaLoader different
     [ 'subdirs', '', '' ],
-    [ 'subdirsstatic', ' LINKTYPE=dynamic', ' LINKTYPE=dynamic' ],
-    [ 'subdirsstatic', ' dynamic', '_dynamic' ],
+    # https://github.com/Perl/perl5/issues/17601
+    # https://rt.cpan.org/Ticket/Display.html?id=115321
+    $^O ne 'MSWin32' ? (
+        [ 'subdirsstatic', ' LINKTYPE=dynamic', ' LINKTYPE=dynamic' ],
+        [ 'subdirsstatic', ' dynamic', '_dynamic' ],
+    ) : (),
     [ 'multi', '', '' ],
-    [ 'staticmulti', ' LINKTYPE=dynamic', ' LINKTYPE=dynamic' ],
-    [ 'staticmulti', ' dynamic', '_dynamic' ],
+    $^O ne 'MSWin32' ? (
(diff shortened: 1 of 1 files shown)


karenetheridge:
It would be good to get to the bottom of these test issues, rather than disabling them.

toddr:
I agree. However, RT-115321 has been open for 4 years now.

Also to Quote @tonycoz:

> %Config provides the information required, but the generated Makefile in this case is generated for LINKTYPE=static, but then run with LINKTYPE=dynamic for Linux that's not such a big deal, if perl is dynamic ccflags includes -fPIC as a matter of course, which just slows the resulting binary, but it is a problem on Win32 due to the way data imports work. I have trouble believing anyone is doing this in the wild ... it might be that the test just isn't practical on Win32, to make it work I think you'd have to use whatever conditional mechanisms the various Win32 makes support.

This has been causing noise on the smokers and I agree with him that turning the test off for a truly corner case like this makes sense until someone comes back and says: `I care about Perl running this way on Windows`. I am dubious that this will ever happen.

Leont:
> It would be good to get to the bottom of these test issues, rather than disabling them.

I think it's understood well enough. It's the sort of thing that is not very portable, and arguably that people shouldn't be doing anyway (but probably someone does)

toddr:
> That said, not passing on LINKTYPE may be a better solution.

@Leont To move this forward, would you be up for providing the alternative pull request?

toddr:
> That said, not passing on LINKTYPE may be a better solution.

@Leont To move this forward, would you be up for providing the alternative pull request?

OR: Can we merge this and you provide the alternative when it is available?

Leont:
> @Leont To move this forward, would you be up for providing the alternative pull request?

> OR: Can we merge this and you provide the alternative when it is available?

Changing this is [trivial](https://metacpan.org/release/ExtUtils-MakeMaker/source/lib/ExtUtils/MM_Unix.pm#L3068), but I haven't figured out if anything would break. I can't think of anything, but this is MakeMaker we're talking about…

toddr:
> Changing this is [trivial](https://metacpan.org/release/ExtUtils-MakeMaker/source/lib/ExtUtils/MM_Unix.pm#L3068), but I haven't figured out if anything would break. I can't think of anything, but this is MakeMaker we're talking about…

And while it's critical, we also get feedback really fast when it breaks something. Given we've thought about it for 4 years. maybe the best option at this point is to try your suggestion and get feedback?

xsawyerx:
> > Changing this is [trivial](https://metacpan.org/release/ExtUtils-MakeMaker/source/lib/ExtUtils/MM_Unix.pm#L3068), but I haven't figured out if anything would break. I can't think of anything, but this is MakeMaker we're talking about…
> 
> And while it's critical, we also get feedback really fast when it breaks something. Given we've thought about it for 4 years. maybe the best option at this point is to try your suggestion and get feedback?

I was going to respond with a similar question. Would it not be useful to put in the suggestion solution and try to receive feedback?

PR #373, opened 2020-10-05
fix PERL_CORE warnings from ExtUtils::CBuilder

### LINKED ISSUE #372: blead: PERL_CORE is set but I can't find your perl source!
While attempting to merge 7.48 into blead, I noticed this TAP output from the perl test suite.

```
$>./TEST ../cpan/ExtUtils-MakeMaker/t/*.t
t/../cpan/ExtUtils-MakeMaker/t/00compile .............. skipped
t/../cpan/ExtUtils-MakeMaker/t/01perl_bugs ............ ok
t/../cpan/ExtUtils-MakeMaker/t/02-xsdynamic ........... PERL_CORE is set but I can't find your perl source!
ok
t/../cpan/ExtUtils-MakeMaker/t/03-xsstatic ............ PERL_CORE is set but I can't find your perl source!
ok
t/../cpan/ExtUtils-MakeMaker/t/arch_check ............. ok
```

haarg:
ExtUtils::CBuilder is used to detect if we have a compiler available.
It has some oddities and doesn't work well in parallel. We run it from a
temporary directory to avoid that. When doing a core build, and in a
temp directory, CBuilder sees PERL_CORE but can't find the perl source
and complains.

If we're running perl core tests, there's no real need to check for a
compiler. We can just assume one is available. Avoid CBuilder's mess and
just return true from our compiler check if running with PERL_CORE.

Fixes #372

### DIFF
--- a/t/lib/MakeMaker/Test/Utils.pm
+++ b/t/lib/MakeMaker/Test/Utils.pm
@@ -359,6 +359,8 @@ Returns true if there is a compiler available for XS builds.
 =cut
 
 sub have_compiler {
+    return 1 if $ENV{PERL_CORE};
+
     my $have_compiler = 0;
 
     in_dir(sub {


Leont:
Looks sensible to me

PR #374, opened 2020-10-09
clean up and improve MIN_PERL_VERSION handling

haarg:
Rather than having special handling for some formats in
MIN_PERL_VERSION, just use version.pm. We have a private implementation
always available even if the real version.pm isn't. This allows versions
like v5.10 or 5.008_001 to be supported.

This also adjusts the warnings and errors to be more consistent and
informative.

This addresses [RT#133489](https://rt.cpan.org/Ticket/Display.html?id=133489)

### DIFF
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -511,50 +511,43 @@ sub new {
 
     check_hints($self);
 
-    if ( defined $self->{MIN_PERL_VERSION}
-          && $self->{MIN_PERL_VERSION} !~ /^v?[\d_\.]+$/ ) {
-      require version;
-      my $normal = eval {
-        local $SIG{__WARN__} = sub {
-            # simulate "use warnings FATAL => 'all'" for vintage perls
-            die @_;
-        };
-        version->new( $self->{MIN_PERL_VERSION} )
-      };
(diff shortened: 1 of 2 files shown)


lib/ExtUtils/MakeMaker.pm
```diff
@@ -511,50 +511,43 @@ sub new {
...
+        if (ref $perl_version) {
+            # assume a version object
+        }
+        else {
+            $perl_version = eval {
+                local $SIG{__WARN__} = sub {
+                    # simulate "use warnings FATAL => 'all'" for vintage perls
+                    die @_;
+                };
+                version->new( $perl_version )->numify;
```
Grinnz, line 525: If it can be used here, `->parse` should be used instead of `->new`.

Grinnz: version isn't being explicitly required anymore, is that intentional?

Grinnz: nevermind the `parse` suggestion as this needs to transparently support version.pm older than 0.77.

haarg: version is loaded by through ExtUtils::MakeMaker::version at the top of the file.

PR #375, opened 2020-10-09
disable strict when evaling hints files

haarg:
Hints files are very commonly written not expecting strict to be
enabled. Prior to PR#367, hints were processed with a 'do', so strict
would not be enabled.

Fixes [RT#133492](https://rt.cpan.org/Ticket/Display.html?id=133492)


toddr (approved):

toddr:
Thanks for submitting this fix!

mohawk2 (approved):

PR #376, opened 2020-10-13
Setup PATH using GitHub API

atoomic:


bingos:
Yeah $GITHUB_PATH is not a Windows environment variable

Needs "Out-File -FilePath $env:GITHUB_PATH -Encoding utf8"

https://github.community/t/migration-from-deprecated-add-path-on-windows/136265

.github/workflows/smoke-windows-strawberryperl.yml
```diff
@@ -26,22 +26,17 @@ jobs:
           perl -MIPC::Cmd -E "say IPC::Cmd::can_run(qq{gcc})"
           perl -MIPC::Cmd -E "say IPC::Cmd::can_run(qq{g++})"
           perl -MIPC::Cmd -E "say IPC::Cmd::can_run(qq{ld})"
-          path C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;%PATH%
+          echo "C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin" ^| Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
bingos, line 29: This should be
```
echo "C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;$env:PATH" ^| Out-File -FilePath $env:GITHUB_PATH -Encoding utf8
```
The Strawberry Perl versions of Mingw need to be ahead of the Chocolately versions

bingos (changes requested):

.github/workflows/smoke-windows-cygwin.yml
```diff
@@ -27,23 +27,17 @@ jobs:
         run: |
           dir c:\tools\cygwin
           path
+          echo "c:\tools\cygwin\bin;c:\tools\cygwin\usr\bin" ^| Out-File -FilePath $env:GITHUB_PATH -Encoding utf8
```
bingos, line 30: Does this need actions/checkout@v2 as well to work?

atoomic: going to give it a try

.github/workflows/smoke-windows-strawberryperl.yml
atoomic: testing with your suggestion via 3aa63b90225c66fec1cddd23b3a108abd99f80a4

.github/workflows/smoke-windows-cygwin.yml
atoomic: this works, I can see failures from `t/02-xsdynamic.t` which is failing under cygwin & strawberry... 
I skipped it using 5688f4c822a715072f3024531e4e387519392cd2 as this seems similar to `t/03-xsstatic.t`

t/02-xsdynamic.t
```diff
@@ -14,6 +14,7 @@ use Test::More;
 plan skip_all => 'Dynaloading not enabled' if !$Config{usedl} or $Config{usedl} ne 'define';
 plan skip_all => "ExtUtils::CBuilder not installed or couldn't find a compiler"
   unless have_compiler();
+plan skip_all => 'Shared perl library' if $Config{useshrplib} eq 'true';
```
Leont, line 17: This change looks unrelated to the rest of this change (and I suspect it's wrong)

bingos: Yeah, I suspect this is hiding a problem

atoomic: I think too that this is hiding another problem

atoomic:
I'm giving up on this one

PR #379, opened 2020-11-09
fix loss of $Config{ccflags} on darwin

tonycoz:
This skips some of the cleanup that EU::MM_Unix::cflags does, it might be better to have an extra_cflags()* method to add the -Wno-error=... instead of using cflags as the original commit did, but this Works For Me.

*or whatever name


haarg:
This doesn't seem like the correct solution. I think the problem is the order. Other OS modules that override cflags call SUPER::cflags first before making their modifications.

tonycoz:
@haarg 's #380 is better, so closing this one.

PR #387, opened 2020-12-26
Improve testing and CI a bit

mohawk2:
This makes the CI do `make disttest NOECHO=` on all platforms, both to exercise the `disttest` functionality a bit more, and to reveal what is happening when it doesn't work.

Similarly, `t/basic.t` is updated so that all runs of the install functionality go via the `check_dummy_inst` function, which will then always check each install systematically in the same way.

### DIFF
--- a/.github/workflows/smoke-windows-cygwin.yml
+++ b/.github/workflows/smoke-windows-cygwin.yml
@@ -47,3 +47,8 @@ jobs:
         run: |
             path c:\tools\cygwin\bin;c:\tools\cygwin\usr\bin
             make test
+      - name: Run Tests
+        shell: cmd
+        run: |
+            path c:\tools\cygwin\bin;c:\tools\cygwin\usr\bin
+            make disttest NOECHO=
--- a/.github/workflows/smoke-windows-strawberryperl.yml
+++ b/.github/workflows/smoke-windows-strawberryperl.yml
@@ -45,3 +45,8 @@ jobs:
         run: |
             path C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;%PATH%
             gmake test
+      - name: Run Tests
+        shell: cmd
+        run: |
+            path C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;%PATH%
+            gmake disttest NOECHO=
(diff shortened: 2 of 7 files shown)


.github/workflows/smoke-windows-strawberryperl.yml
```diff
@@ -45,3 +45,8 @@ jobs:
         run: |
             path C:\Strawberry\c\bin;C:\Strawberry\perl\site\bin;C:\Strawberry\perl\bin;%PATH%
             gmake test
+      - name: Run Tests
```
haarg, line 48: This too.

t/basic.t
```diff
@@ -186,64 +186,48 @@ open $fh, '>', $makefile;
...
-    ok( open(PERLLOCAL, $files->{'perllocal.pod'} ) ) ||
-        diag("Can't open $files->{'perllocal.pod'}: $!");
-    { local $/;
-      unlike(<PERLLOCAL>, qr/other/, 'DESTDIR should not appear in perllocal');
-    }
-    close PERLLOCAL;
+    my $files = check_dummy_inst('other', 'PREFIX= DESTDIR=other', 'with DESTDIR');
+    eval {
+        unlike slurp($files->{'perllocal.pod'}), qr/other/, 'DESTDIR should not appear in perllocal';
+    };
```
haarg, line 229: Seems like if there is an exception here, it wouldn't run the test function, which would throw off the text count for the rest of the file.

.github/workflows/smoke-windows-cygwin.yml
```diff
@@ -47,3 +47,8 @@ jobs:
         run: |
             path c:\tools\cygwin\bin;c:\tools\cygwin\usr\bin
             make test
+      - name: Run Tests
```
haarg, line 50: This should probably have a distinct name.

haarg (changes requested):

.github/workflows/smoke-windows-strawberryperl.yml
mohawk2: Agreed, thanks.

t/basic.t
mohawk2: Agreed, adjusted.

.github/workflows/smoke-windows-cygwin.yml
mohawk2: Adjusted.

PR #390, opened 2021-03-20
Fix warning from undefined thislib

atoomic:
```
===Use of uninitialized value $thislib in substitution (s///) at ../lib/ExtUtils/Liblist/Kid.pm line 72, <DATA> line 53.
Use of uninitialized value $thislib in pattern match (m//) at ../lib/ExtUtils/Liblist/Kid.pm line 99, <DATA> line 53.
Use of uninitialized value $thislib in substitution (s///) at ../lib/ExtUtils/Liblist/Kid.pm line 106, <DATA> line 53.
Use of uninitialized value $thislib in concatenation (.) or string at ../lib/ExtUtils/Liblist/Kid.pm line 114,
```

### DIFF
--- a/lib/ExtUtils/Liblist/Kid.pm
+++ b/lib/ExtUtils/Liblist/Kid.pm
@@ -96,6 +96,8 @@ sub _unix_os2_ext {
             next;
         }
 
+        next unless defined $thislib;
+
         if ( $thislib =~ m!^-Wl,! ) {
             push( @extralibs,  $thislib );
             push( @ldloadlibs, $thislib );


lib/ExtUtils/Liblist/Kid.pm
```diff
@@ -96,6 +96,8 @@ sub _unix_os2_ext {
             next;
         }
 
+        next unless defined $thislib;
```
Grinnz, line 99: This doesn't make a lot of sense to me. The only possible way for $thislib to be undefined here is if quotewords returned an undefined argument (in which case the first substitution would warn before this line) or if it's assigned undef from catdir (does not seem possible) or `qq{"$thislib"}` (not possible).

Grinnz: Indeed your first warning is from the initial substitution which means quotewords returned an `undef` for some reason.

Leont:
I'm pretty sure this fix is wrong, and would propose closing it.

Can you instead post an issue first describing how you encountered this, so that we can analyze what's really going on?

Leont:
> I'm pretty sure this fix is wrong, and would propose closing it.
> 
> Can you instead post an issue first describing how you encountered this, so that we can analyze what's really going on?

After some experimenting I discovered that if `$potential_libs` equals `" "`, `quotewords` will return the list `('', undef)`. I'm guessing that's how this is triggered. This issue was introduced in d21fe269f.

Possibly we should be using `shellwords` instead, but we should check to be sure if that doesn't break anything (this is LibList, the odds are non-trivial). @mohawk2: is there a specific reason you used `quotewords` instead of `shellwords`?

atoomic:
let's close it, this was a quick n dirty suggestion to fix the warning from blead in 5.33.8

Leont:
> let's close it, this was a quick n dirty suggestion to fix the warning from blead in 5.33.8

MakeMaker is not the kind of codebase where quick and dirty suggestions are helpful, it already contains more dirt than it can carry. Next time please file a bug report instead.

wchristian:
As someone who worked on this and spent hours trawling and tracing through this code with the realtime debugger, and added tests for everything:

Please always make sure that any change is done with with plenty time invested so you know exactly what you're doing and includes copious amounts of tests verifying circumstances before and after your change.

This is not a place where anything quick, nevermind dirty, can be done.

atoomic:
reported as #391 and fixed as #392

PR #395, opened 2021-05-17
comparing inodes numerically is unsafe

tonycoz:
On a 32-bit system with 64-bit inodes, recent-ish versions of perl
will return very large inode numbers as strings, and their numeric
value may be too large to be exactly represented as an NV.

This means comparing such inodes numerically may return them as equal
even when the actual inode number is different.

See https://github.com/Perl/perl5/pull/18788 for a case where this
has caused a problem.


rokuyama:
Hi,

I think `lib/File/Copy.pm` should be fixed also:
https://gist.github.com/rokuyama/f8949a8bc552cb7c7b9d797c4ebb575a

Thanks,
rin

tonycoz:
> I think `lib/File/Copy.pm` should be fixed also:

Fixed in 76d2194af (on perl/perl5)

rokuyama:
Thank you for your quick response!

Issue #290, opened 2017-02-16
Eliminate legacy classes?


Leont:
We have a number of classes for supporting operating systems that we can't reasonably claim to support anymore. These include `MM_Win95`, `MM_DOS`, `MM_NW5` and `MM_OS2`. These do still cause maintenance time. These should probably get the same treatment that `MM_MacOS` has already received (replaced by a fatal stub).

Leont:
This would for example make #91 a bit easier.

mohawk2:
Zapping DOS support means we could also remove this snippet from `MM_Unix`:

```
    push @m, "\t", q{$(NOECHO) $(PERL) "$(INSTALLSCRIPT)/fixpmain"
} if (defined (&Dos::UseLFN) && Dos::UseLFN()==0);
```

PR #396, opened 2021-06-18
Retrofit haikuports patchset to support haiku as a separated platform from BeOS

thibaultduponchelle:
Hello 😀 

I propose this change to retrofit the portions of haikuports patch related to EUMM. See [haikuports perl patchset](https://github.com/haikuports/haikuports/blob/master/dev-lang/perl/patches/perl-5.32.1.patchset#L186)

Haiku has now some differences versus BeOS (filesystem hierarchy conventions...) that justifies this new "platform variant".

I tested it with perl 5.34.0 but I'm not a haiku expert and the code is not mine but from @olta

@bingos How do you feel about this change ?

### DIFF
--- a/lib/ExtUtils/MM.pm
+++ b/lib/ExtUtils/MM.pm
@@ -61,7 +61,8 @@ if( $^O eq 'MSWin32' ) {
 $Is{UWIN}   = $^O =~ /^uwin(-nt)?$/;
 $Is{Cygwin} = $^O eq 'cygwin';
 $Is{NW5}    = $Config{osname} eq 'NetWare';  # intentional
-$Is{BeOS}   = ($^O =~ /beos/i or $^O eq 'haiku');
+$Is{BeOS}   = $^O =~ /beos/i;
+$Is{Haiku}  = $^O eq 'haiku';
 $Is{DOS}    = $^O eq 'dos';
 if( $Is{NW5} ) {
     $^O = 'NetWare';
--- a/lib/ExtUtils/MM_Haiku.pm
+++ b/lib/ExtUtils/MM_Haiku.pm
@@ -1,0 +1,63 @@
+package ExtUtils::MM_Haiku;
+
+use strict;
+use warnings;
+
+=head1 NAME
+
+ExtUtils::MM_Haiku - methods to override UN*X behaviour in ExtUtils::MakeMaker
+
+=head1 SYNOPSIS
+
+ use ExtUtils::MM_Haiku;    # Done internally by ExtUtils::MakeMaker if needed
+
+=head1 DESCRIPTION
+
+See ExtUtils::MM_Unix for a documentation of the methods provided
+there. This package overrides the implementation of these methods, not
+the semantics.
+
+=over 4
+
+=cut
+
+use ExtUtils::MakeMaker::Config;
+use File::Spec;
+use ExtUtils::MM_Unix;
+
+our @ISA = qw( ExtUtils::MM_Unix );
+our $VERSION = '7.63_02';
+$VERSION =~ tr/_//d;
+
+
+sub os_flavor {
+    return('Haiku');
+}
+
+sub init_main {
+    my $self = shift;
+
+    # switch to vendor directories if requested.
+    if ($ENV{'HAIKU_USE_VENDOR_DIRECTORIES'}) {
+        $self->{INSTALLDIRS} ||= 'vendor';
+    }
+
+    $self->SUPER::init_main();
+}
+
+sub init_others {
+    my $self = shift;
+
+    $self->SUPER::init_others();
+
+    # Don't use run-time paths for libraries required by dynamic
+    # modules on Haiku, as that wouldn't work should a library be moved
+    # (for instance because the package has been activated somewhere else).
+    $self->{LD_RUN_PATH} = "";
+
+    return;
+}
+
+1;
+__END__
+


lib/ExtUtils/MM_Haiku.pm
```diff
@@ -0,0 +1,81 @@
...
+=over 4
+
+=cut
+
+use ExtUtils::MakeMaker::Config;
+use File::Spec;
+require ExtUtils::MM_Any;
+require ExtUtils::MM_Unix;
+
+our @ISA = qw( ExtUtils::MM_Any ExtUtils::MM_Unix );
```
Leont, line 28: No. Just no.

I really don't think we should be deriving form `ExtUtils::MM_Any` here. Let's not propagate that headache.

lib/ExtUtils/MM.pm
```diff
@@ -61,7 +61,8 @@ if( $^O eq 'MSWin32' ) {
 $Is{UWIN}   = $^O =~ /^uwin(-nt)?$/;
 $Is{Cygwin} = $^O eq 'cygwin';
 $Is{NW5}    = $Config{osname} eq 'NetWare';  # intentional
-$Is{BeOS}   = ($^O =~ /beos/i or $^O eq 'haiku');
+$Is{BeOS}   = $^O =~ /beos/i;
```
Leont, line 64: Does keeping `$Is{BeOS}` even make sense?

lib/ExtUtils/MM_Haiku.pm
```diff
@@ -0,0 +1,81 @@
...
+    # but due to the way how Haiku's package management works, PERLPREFIX
+    # and VENDORPREFIX are not writable at all (as they're being populated
+    # from installed packages via package-fs). SITEPREFIX, however needs to
+    # be set to a path which can be written to (since site packages are
+    # expected to be installed "manually") - so we make sure it points
+    # to a 'non-packaged'-folder:
+    my $prefixGiven = $self->{PREFIX};
+	$self->SUPER::init_INSTALL_from_PREFIX();
+	if ($prefixGiven) {
+        $self->{SITEPREFIX} = '$(PREFIX)/non-packaged';
```
Leont, line 62: This doesn't make sense to me. I'm suspecting they're using this the wrong way.

thibaultduponchelle: I have the same hesitation about this line

lib/ExtUtils/MM_Haiku.pm
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+sub init_others {
+    my $self = shift;
+
+	$self->SUPER::init_others();
```
Leont: The whitespace is inconsistent in various places.

lib/ExtUtils/MM_Haiku.pm
thibaultduponchelle: Fixed

lib/ExtUtils/MM_Haiku.pm
thibaultduponchelle: ouch sorry for that

thibaultduponchelle: fixed

lib/ExtUtils/MM.pm
thibaultduponchelle: I don't know, I should ask

lib/ExtUtils/MM_Haiku.pm
```diff
@@ -0,0 +1,81 @@
+package ExtUtils::MM_Haiku;
+
+use strict;
+
```
Leont, line 4: `use warnings;`?

lib/ExtUtils/MM_Haiku.pm
```diff
@@ -0,0 +1,81 @@
...
+there. This package overrides the implementation of these methods, not
+the semantics.
+
+=over 4
+
+=cut
+
+use ExtUtils::MakeMaker::Config;
+use File::Spec;
+require ExtUtils::MM_Any;
```
Leont, line 25: These two can just be a `use ExtUtils::MM_Unix;`

thibaultduponchelle: I fixed it, but notes in my update I kept the `require` instead of `use` (is it good?)

lib/ExtUtils/MM_Haiku.pm
thibaultduponchelle: fixed thx

lib/ExtUtils/MM.pm
Leont: I suppose we should keep it for now, and re-evaluate when we look at  #290

Leont:
This should also adapt `t/MM_BeOS.t`

lib/ExtUtils/MM_Haiku.pm
Leont: It's not wrong, just a little silly

thibaultduponchelle:
I removed a portion of the patch since it should be fixed at perl5 compilation level: if you use -Dprefix... set also siteprefix and you're all good, and if not just use default `./Configure -des` that will set correctly the site prefix to a non-packaged and writable directory :+1:

I changed the `require` to `use`. 

I kept the first function to use vendor on env var, it seems intended to haiku packagers (can be emulated easily for one package with `perl Makefile.PL INSTALLDIRS=vendor`) and I also kept the LD_RUN_PATH patch that seems the most interesting (even if I don't understand it 100%).

Concerning the BeOS support, I'm asking a status to some people from Haiku community to have a first idea, but maybe we can keep this part out of this PR?

lib/ExtUtils/MM_Haiku.pm
thibaultduponchelle: I removed this portion of the patch

thibaultduponchelle:
Having continued to work on this topic, I'm not 100% convinced about this change, so I will close it for now, waiting to know what is really needed and what is not.

(Thank you for the code review)

PR #398, opened 2021-06-26
Don't use canned libpth values

Leont:
On many platforms, these can change with compiler upgrades. Instead of using the values that were valid when perl was configured, this will ask the compiler for the valid values if possible.


xenu:
tbh I don't understand why EU::MM verifies the existence of libraries at all. It seems that the only purpose of that is to cause build errors.

Leont:
> tbh I don't understand why EU::MM verifies the existence of libraries at all. It seems that the only purpose of that is to cause build errors.

I also think it's a misfeature, but people depend on it. In particular they depend on giving multiple options, and MakeMaker picking the first one that would succeed.

mohawk2 (approved):

PR #399, opened 2021-06-30
Describe CCFLAGS’ default.

FGasper:


Leont:
Same is true for `OPTIMIZE`, actually. I had expected there to be more non-path %Config defaulting arguments, but apparently not so much.

PR #401, opened 2021-07-23
Add CPPRUN variable

Leont:
This adds an `CPPRUN` variable as suggested in [#137912](https://rt.cpan.org/Ticket/Display.html?id=137912).


mohawk2 (approved):
Looks good! Probably worth documenting though?

Leont:
> Looks good! Probably worth documenting though?

Fair. Added it to the documented list of such variables.

Issue #402, opened 2021-09-16
LC_RPATH not added when compiling XS modules on macOS


hakonhagland:
## Introduction

I did some research to try to better understand the way shared libraries are generated and used on macOS. I include it here as an introduction. See below for the actual issue with `ExtUtils::MakeMaker`

On Linux it is enough to set the environment variable `LC_RUN_PATH` when compiling a shared library. This will make the linker include the run path given in the `LC_RUN_PATH` variable into the shared library. On macOS it is more complicated. Each library has a header providing an install name (`LC_ID_DYLIB`) of the library. If other libraries should be able to find this library through their own run time path (rpath) the install name of the current library must begin with the special string `@rpath`. Secondly, if this library is trying to load another library using its rpath the header section of this library called `LC_LOAD_DYLIB` must include the rpath install name of the other library.  This means that the name in the `LC_LOAD_DYLIB` section must start with the special string `@rpath`. Thirdly, this library must include an `LC_RPATH` section with the run time paths to the libraries (the ones whose names starts with `@rpath`) given in the `LC_LOAD_DYLIB` section.

When linking a shared library on macOS (using here `cc` = `clang` as the most relevant example), 
- the switch `-install_name` can be used to set the `LC_ID_DYLIB` section for the generated shared object,
- the switch `-rpath` or `-Wl,-rpath,` can be used to set the `LC_RPATH` section,
- and the environment variable `LD_RUN_PATH` can be used to set the `@rpath` prefix of library names in the `LC_LOAD_DYLIB` section

See [this](https://matthew-brett.github.io/docosx/mac_runtime_link.html) link for more information. 

Here is an example: I have the following files on my mac (Big Sur 11.5.2) :
```
$ tree .
.
├── mylib1
│   ├── libmylib1.dylib
│   ├── mylib1.c
│   └── mylib1.h
├── mylib2
│   ├── libmylib2.dylib
│   ├── mylib2.c
│   └── mylib2.h
└── program
    ├── main
    └── main.c

3 directories, 8 files
```
and the file contents are:
- **mylib1.c** : 
```
#include <stdio.h>
#include "mylib1.h"
void foo() {
    printf( "Hello from foo()\n");
}
```
- **mylib2.c** : 
```
#include <stdio.h>
#include "mylib1.h"
#include "mylib2.h"
void bar() {
    printf( "Hello from bar()\n");
    foo();
}
```  
- **main.c**:
```
[...]

PR #403, opened 2021-09-16
Add -rpath when compiling XS modules on macOS

hakonhagland:
This change fixes issue #402 on my machine.


Leont:
Can you please change the commit title to something that actually describe what it does?

hakonhagland:
@Leont I have updated the title to something more descriptive.

haarg:
The important thing to fix is the subject of the actual commit, not the title of the PR.

hakonhagland:
@haarg I have update the commit message with the same title

mohawk2 (approved):
Looks good to me, and the CI likes it.

mohawk2:
@hakonhagland Could you add a test that breaks on MacOS without this change? That way we can permanently capture the benefit.

PR #405, opened 2021-11-26
Add $(LDFLAGS) when linking binary modules

kanavin:
LDFLAGS as environment variable is commonly used to pass global linker
flags in distribution builds to specific components build systems.

E.g. Yocto project sets --hash-style --debug-prefix-map and other entries in there.

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -1092,7 +1092,7 @@ sub xs_make_dynamic_lib {
     }
 
     push @m, sprintf <<'MAKE', $ld_run_path_shell, $ldrun, $dlsyms_arg, $ldfrom, $self->xs_obj_opt('$@'), $libs, $exportlist;
-	%s$(LD) %s $(LDDLFLAGS) %s %s $(OTHERLDFLAGS) %s $(MYEXTLIB) \
+	%s$(LD) %s $(LDDLFLAGS) %s %s $(LDFLAGS) $(OTHERLDFLAGS) %s $(MYEXTLIB) \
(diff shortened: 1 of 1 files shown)


mohawk2:
This seems sensible on the face of it, and the CI passes.

Leont:
LDFLAGS is "the flags that the perl binary itself will be linked with", in particular that means the flags are suitable for an executable instead of a loadable object. The latter is done using `LDDLFLAGS` which largely but not fully overlaps with `LDFLAGS`.

Also, this has absolutely nothing to do with the `LDFLAGS` environment variable, it's the `ldflags` configuration variable.

IMNSHO this is wrong, and should be reverted.

kanavin:
> LDFLAGS is "the flags that the perl binary itself will be linked with", in particular that means the flags are suitable for an executable instead of a loadable object. The latter is done using `LDDLFLAGS` which largely but not fully overlaps with `LDFLAGS`.
> 
> Also, this has absolutely nothing to do with the `LDFLAGS` environment variable, it's the `ldflags` configuration variable.
> 
> IMNSHO this is wrong, and should be reverted.

Yocto project has been carrying this patch for many years. Without it, what we set in LDFLAGS environment variable does not propagate to things that are built with MakeMaker, and we get errors in binary QA checks. If this is wrong, please suggest the correct way to do it.

Leont:
> Without it, what we set in LDFLAGS environment variable does not propagate to things that are built with MakeMaker, and we get errors in binary QA checks.

Given that our Makefile set their own LDFLAGS, that sounds unlikely. Or are you explicitly passing it into MakeMaker?

PR #406, opened 2021-11-27
Add $(LDFLAGS) when linking binary modules

kanavin:
LDFLAGS as environment variable is commonly used to pass global linker
flags in distribution builds to specific components build systems.

E.g. Yocto project sets --hash-style --debug-prefix-map and other entries in there.

Signed-off-by: Alexander Kanavin <alex@linutronix.de>

Note: this was merged in https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/405, then reverted under the claim that it's not the correct way to pass items in LDFLAGS env var, however no alternative was offered. Therefore I have to re-submit, and ask for that alternative. @Leont FYI

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -1092,7 +1092,7 @@ sub xs_make_dynamic_lib {
     }
 
     push @m, sprintf <<'MAKE', $ld_run_path_shell, $ldrun, $dlsyms_arg, $ldfrom, $self->xs_obj_opt('$@'), $libs, $exportlist;
-	%s$(LD) %s $(LDDLFLAGS) %s %s $(OTHERLDFLAGS) %s $(MYEXTLIB) \
+	%s$(LD) %s $(LDDLFLAGS) %s %s $(LDFLAGS) $(OTHERLDFLAGS) %s $(MYEXTLIB) \
 	  $(PERL_ARCHIVE) %s $(PERL_ARCHIVE_AFTER) %s \
 	  $(INST_DYNAMIC_FIX)
 	$(CHMOD) $(PERM_RWX) $@


kanavin:
Just to repeat, Yocto project has been carrying this patch for many years. Without it, what we set in LDFLAGS environment variable does not propagate to things that are built with MakeMaker, and we get errors in binary QA checks. If this is wrong, please suggest the correct way to do it.

Leont:
> Without it, what we set in LDFLAGS environment variable does not propagate to things that are built with MakeMaker, and we get errors in binary QA checks.

Given that our Makefile set its own LDFLAGS (that takes precedence over any environmental variable), that sounds unlikely. Or are you explicitly passing it into MakeMaker?

kanavin:
> Given that our Makefile set its own LDFLAGS (that takes precedence over any environmental variable), that sounds unlikely. Or are you explicitly passing it into MakeMaker?

I checked what is happening in our builds. We do not use perl's own build system, as it is entirely unsuited to cross compilation - instead we use perl-cross from http://arsv.github.io/perl-cross/

perl-cross writes LDFLAGS from environment into perl config files as ldflags, and (I think) this is picked up by makemaker and written into Makefiles produced by it.  So far so good, but makemaker's makefiles do not actually pass that LDFLAGS setting to the linker (or use it in any other way). The change in this PR adds that missing bit.

Leont:
> I checked what is happening in our builds. We do not use perl's own build system, as it is entirely unsuited to cross compilation - instead we use perl-cross from http://arsv.github.io/perl-cross/

Ah! That is crucial information that was previously missing.

> perl-cross writes LDFLAGS from environment into perl config files as ldflags, and (I think) this is picked up by makemaker and written into Makefiles produced by it. So far so good, but makemaker's makefiles do not actually pass that LDFLAGS setting to the linker (or use it in any other way). The change in this PR adds that missing bit.

That sounds rather suboptimal. It should probably be adding the environment variable `LDFLAGS` not just to the `ldflags` configuration variable but also to `lddlflags`.

kanavin:
> That sounds rather suboptimal. It should probably be adding the environment variable `LDFLAGS` not just to the `ldflags` configuration variable but also to `lddlflags`.

There is some logic in perl-cross that seem to do just that, but with some kind of filtering that I do not fully understand:
https://github.com/arsv/perl-cross/blob/master/cnf/configure_tool.sh#L189
I'll take a closer look, as either it doesn't work at all or filters out important items - it probably needs to be replaced by a complete copy from one to the other, no questions asked.

kanavin:
I can confirm that with the perl-cross 'filter' removed, the flags are correctly passed all the way to the linker:
```
Checking which flags from $ldflags to move to $lddlflags
    added -Wl,-O1
    added -Wl,--hash-style=gnu
    added -Wl,--as-needed
    added -fmacro-prefix-map=/home/alex/development/poky/build-64-alt/tmp/work/core2-64-poky-linux/perl/5.34.0-r0=/usr/src/debug/perl/5.34.0-r0
    added -fdebug-prefix-map=/home/alex/development/poky/build-64-alt/tmp/work/core2-64-poky-linux/perl/5.34.0-r0=/usr/src/debug/perl/5.34.0-r0
    added -fdebug-prefix-map=/home/alex/development/poky/build-64-alt/tmp/work/core2-64-poky-linux/perl/5.34.0-r0/recipe-sysroot=
    added -fdebug-prefix-map=/home/alex/development/poky/build-64-alt/tmp/work/core2-64-poky-linux/perl/5.34.0-r0/recipe-sysroot-native=
    added -Wl,-z,relro,-z,now
```

I'll take this up with perl-cross upstream, thanks for the help!

PR #411, opened 2021-12-29
cross-compiling needs a real compiler check

fperrad:
the short cut introduced by the commit https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/5e55318a97616596447c336b9cb331a8dfdc62d5 must depend on `$Config{'usecrosscompile'}`.


mohawk2:
This seems sensible.

Leont:
Doesn't that reintroduce the issue that 5e55318 sought to fix for cross-compiling perls?

fperrad:
after comments from @Leont , I rewrite the patch in a saner way : the short cut always applies when `PERL_CORE`.

mohawk2:
Other code detects cross-compilation using `if (defined $::Cross::platform)` - should this as well?

PR #414, opened 2022-03-14
fix typos using WriteMakeFile rather than WriteMakefile

haarg:


demerphq (approved):
Does this patch need a corresponding version bump?

demerphq:
BTW, this resolves https://rt.cpan.org/Ticket/Display.html?id=141753

mohawk2 (approved):

haarg:
Version bumping in this repo is handled in the release process, not with the individual changes.

PR #415, opened 2022-05-03
Makefile overwrites /dev/null

sw0x2A:
When building LaTeXML from CPAN, the Makefile overwrites /dev/null under some circumstances. It is in the clean section and after some debugging I found this comes from [Makemaker](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/blob/master/lib/ExtUtils/MM_Any.pm#L777).

`$(MV) $(FIRST_MAKEFILE) $(MAKEFILE_OLD) $(DEV_NULL)`

This line moves both FIRST_MAKEFILE and MAKEFILE_OLD to DEV_NULL and replaces it. [The comment above the block](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/blob/master/lib/ExtUtils/MM_Any.pm#L774) let me assume that FIRST_MAKEFILE should be renamed to MAKEFILE_OLD. The DEV_NULL must be a mistake but I am a bit surprised because this line has not been changed in 18 years, so this bug is quite mature.

### DIFF
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -774,7 +774,7 @@ clean :: clean_subdirs
     # Leave Makefile.old around for realclean
     push @m, <<'MAKE';
 	  $(NOECHO) $(RM_F) $(MAKEFILE_OLD)
-	- $(MV) $(FIRST_MAKEFILE) $(MAKEFILE_OLD) $(DEV_NULL)
+	- $(MV) $(FIRST_MAKEFILE) $(MAKEFILE_OLD)
 MAKE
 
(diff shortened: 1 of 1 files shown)


haarg:
The normal content of `DEV_NULL` is [`> /dev/null 2>&1`](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/blob/master/lib/ExtUtils/MM_Any.pm#L2392), so this diagnosis doesn't seem quite correct.

What type of system are you working with? Can you check how `DEV_NULL` is set in the generated `Makefile`?

sw0x2A:
Hmm. I don't have access to that system for a while and can't follow up. Obviously, my PR doesn't make much sense, therefore closing it.

PR #418, opened 2022-05-16
use Test::More rather than Test.pm in FAQ example

haarg:


Leont (approved):

lib/ExtUtils/MakeMaker/FAQ.pod
```diff
@@ -536,8 +536,7 @@ And of course a very basic test:
 
   t/cool.t:
   --------
-  use Test;
-  BEGIN { plan tests => 1 };
+  use Test::Mote tests => 1;
```
mohawk2, line 539: Maybe `Test::More` instead of `Test::Mote`? :-)

PR #419, opened 2022-05-31
Only include existent directories in libpath

Leont:
Otherwise, one can try to open non-existent directories, leading to warnings about invalid dirhandles.

This fixes https://github.com/Perl/perl5/issues/19794

### DIFF
--- a/lib/ExtUtils/Liblist/Kid.pm
+++ b/lib/ExtUtils/Liblist/Kid.pm
@@ -62,6 +62,7 @@ sub _unix_os2_ext {
 		chomp(my @incpath = grep s/^ //, grep { /^#include </ .. /^End of search / } `$Config{cc} -E -v - </dev/null 2>&1 >/dev/null`);
 		unshift @libpath, map { s{/include[^/]*}{/lib}; $_ } @incpath
 	}
+	@libpath = grep -d, @libpath;
 
     if ( $^O eq 'darwin' or $^O eq 'next' )  {
         # 'escape' Mach-O ld -framework and -F flags, so they aren't dropped later on


mohawk2:
This does seem like a good idea.

Sidebar: the Cygwin CI failure is saying this:

```
fatal: unsafe repository ('/cygdrive/d/a/ExtUtils-MakeMaker/ExtUtils-MakeMaker' is owned by someone else)
To add an exception for this directory, call:
	git config --global --add safe.directory /cygdrive/d/a/ExtUtils-MakeMaker/ExtUtils-MakeMaker
```

That's in the stage called "Check out again, using Cygwin git, to reset file permissions", which suggests that it's not achieving its aim.

haarg (approved):

Leont:
> Sidebar: the Cygwin CI failure is saying this:

Yeah, it probably needs a `git config --global --add safe.directory *`

Leont:
> Unfortunately this does not appear to have solved the problem.

Then I guess all of `@libpath` needs to be filtered on `-d`

mohawk2:
Apart from a cross-build scenario (and maybe not even then), can anyone see a good reason to allow non-existent directories in any of the "path" arrays?

Leont:
Pushed a new version that should resolve the issues you're seeing.

jkeenan:
> Pushed a new version that should resolve the issues you're seeing.

Will test this out tomorrow. Thanks.

jkeenan:
> > > Pushed a new version that should resolve the issues you're seeing.
> > 
> > 
> > Will test this out tomorrow. Thanks.
> 
> This clears up the Perl-build-time-warnings I reported in [Perl/perl5#19794](https://github.com/Perl/perl5/issues/19794). I took your patch and applied it to a batch in Perl/perl5 and compiled with 2 different compilers on 2 different OSes, to good effect.
> 
> Thank you very much. Jim Keenan

The current state of this p.r. *may* need some provision to deal with CI failures on cygwin.  Otherwise, can we get it ready for a CPAN release?

PR #424, opened 2022-12-23
use a poor mans tempdir logic for parallel tests

demerphq:
In core we want to run the tests in parallel. The tests in EUMM are not parallel safe as they all use the same directory for BFD and then the teardown from one affects the other, etc. This adds a poor mans temporary directory so each test invocation gets its own copy of the BFD directory. This has been tested in core builds already, and migrated to the master repo for EUMM from there.

This should fix https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/425
See also: https://github.com/Perl/perl5/pull/20643

### DIFF
--- a/t/INST.t
+++ b/t/INST.t
@@ -76,7 +76,7 @@ is( !!$mm->{PERL_CORE}, !!$ENV{PERL_CORE}, 'PERL_CORE' );
 
 my($perl_src, $mm_perl_src);
 if( $ENV{PERL_CORE} ) {
-    $perl_src = File::Spec->catdir($Updir, $Updir, $Updir, $Updir, $Updir);
+    $perl_src = File::Spec->catdir($Updir, $Updir, $Updir, $Updir, $Updir, $Updir);
     $perl_src = File::Spec->canonpath($perl_src);
     $mm_perl_src = File::Spec->canonpath($mm->{PERL_SRC});
 }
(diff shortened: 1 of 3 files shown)


demerphq:
It isnt clear to me if this patch breaks Cygwin, or if it is just broken. :-(

karenetheridge:
If it helps, a copy of File::Temp comes bundled with EUMM and can be used in tests without introducing a dependency.

demerphq:
On Fri, 23 Dec 2022, 20:03 Karen Etheridge, ***@***.***>
wrote:

> If it helps, a copy of File::Temp comes bundled with EUMM and can be used
> in tests without introducing a dependency.
>

Well, I wanted to make the change as minimal as possible, and the setup
code is called by test logic that blocks XS which breaks use of File::Temp
(via it loading IO related XS code). FWIW when I first tried to fix this
issue I tried to do this with File::Temp, but it broke, i dont hand roll
code like this for fun ;-). The EUMM tests are twisty enough I didn't want
risk breaking the tests by working around this by preloading the XS code
before the block is put in place, and i really wanted to fix this without
having to modify each of the test files...

All of this could be worked around by more serious changes to each test
that uses the BFD code, but that is a lot of work for little return IMO,
this is test logic for code that has to work in extreme conditions, eg
without XS support. The simpler it is the less can go wrong.

Cheers
Yves

PR #426, opened 2022-12-24
Setup/BFD - use 0..9 not -9 as characters in tempdir

demerphq:
It doesn't matter much, in sense that the code works fine, but 0..9 is what I intended. :-)


demerphq:
Sorry about that. Mixed up charclass and range notation, and since it was digits perl didnt notice. Doesn't change the result, it still creates temp files ok, it just doesn't do it quite like i had intended. Sigh. late night hacks.

PR #432, opened 2023-02-19
fix MIN_PERL_VERSION for perl versions with underscores

haarg:
version->numify throws a warning when numifying alpha versions. numify throws warnings based on the caller's lexical warnings. We don't care. We just want the number.

### DIFF
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -525,7 +525,10 @@ sub new {
                     # simulate "use warnings FATAL => 'all'" for vintage perls
                     die @_;
                 };
-                version->new( $perl_version )->numify;
+                my $v = version->new($perl_version);
+                # we care about parse issues, not numify warnings
+                no warnings;
+                $v->numify;
             };
             $perl_version =~ tr/_//d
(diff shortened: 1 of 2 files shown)


lib/ExtUtils/MakeMaker.pm
```diff
@@ -525,7 +525,10 @@ sub new {
                     # simulate "use warnings FATAL => 'all'" for vintage perls
                     die @_;
                 };
-                version->new( $perl_version )->numify;
+                my $v = version->new($perl_version);
+                # we care about parse issues, not numify warnings
+                no warnings;
```
Leont, line 530: `no warnings "numeric"` may be better here.

haarg: It would have to be `no warnings $] > 5.015 ? 'version' : 'numeric';`. I didn't think it would be worth it. There is really nothing `->numify` could ever warn about that we would ever care about, given that it already succeeded in parsing the version.

Leont: I see, fair enough

Leont (approved):

Issue #434, opened 2023-02-28
Existing META.yml is given too much priority when generating MYMETA.yml and etc, resulting in lost License data


demerphq:
I was working on a BBC ticket for perl, https://github.com/Perl/perl5/issues/20864 and I noticed that it complained that the LICENSE value from the Makefile.PL was invalid.  The value is/was "perl_5", which as far as I can tell *is* a valid value.

So I started digging into it. Eventually I found my way to ExtUtils::MM_Any::_fix_metadata_before_conversion().  I instrumented the function to dump out its arguments and cluck a stacktrace when it was called, and it seems to get called twice. Once from metafile_target(), and once from write_mymeta(). 

In the first call the license value is recognized as correct, and things look (to me) to be correct. On the second call however things look much different, actually downright strange, and the license is not recognized and turns into "unknown".  The second call seems to be responsible for generating the MYMETA.yaml and MYMETA.json files and both say the license is "unknown". Furthermore they both say this:

```
  "generated_by" : "ExtUtils::MakeMaker version 6.44, CPAN::Meta::Converter version 2.150010",
```

note that this is from a perl that was built from blead a day ago, with ExtUtils::MakeMaker 7.66.

Here is the instrumented output:
```
~/git_tree/Slay-Makefile$ perl Makefile.PL 
$VAR1 = {
          'abstract' => 'Wrapper to Slay::Maker that reads the rules from a file',
          'author' => [
                        'Slaven Rezic <srezic@cpan.org>',
                        'Mark Nodine <mnodine@alum.mit.edu>'
                      ],
          'dynamic_config' => 1,
          'generated_by' => 'ExtUtils::MakeMaker version 7.66',
          'license' => [
                         'perl_5'
                       ],
          'meta-spec' => {
                           'url' => 'https://metacpan.org/pod/CPAN::Meta::Spec',
                           'version' => 2
                         },
          'name' => 'Slay-Makefile',
          'no_index' => {
                          'directory' => [
                                           't',
                                           'inc'
                                         ]
                        },
          'prereqs' => {
                         'build' => {
                                      'requires' => {
                                                      'ExtUtils::MakeMaker' => 0
                                                    }
                                    },
                         'configure' => {
                                          'requires' => {
                                                          'ExtUtils::MakeMaker' => 0
                                                        }
                                        },
                         'runtime' => {
                                        'requires' => {
                                                        'File::Copy::Recursive' => '0',
[...]

PR #436, opened 2023-02-28
Fix strange warning from existing META.yml license info conflicting with Makefile.PL

demerphq:
See https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/434 for details of an example of this.

This patch keeps track of the generated metadata from the WriteMakefile() command, and then later when we load the data from META.yml we notice if the license data differs between the two. 

This avoids the incorrect warning about an invalid license, and allows us to point out when there is a discrepancy so developers can fix it.


lib/ExtUtils/MM_Any.pm
```diff
@@ -1671,7 +1687,7 @@ sub _mymeta_from_meta {
     # rolled their own tarball rather than using "make dist".
     if ($meta->{generated_by} &&
         $meta->{generated_by} =~ /ExtUtils::MakeMaker version ([\d\._]+)/) {
-        my $eummv = do { no warnings; $1+0; };
+        my $eummv = do { no warnings; eval "$1" + 0; }; # deal with underbars gracefully
```
Leont, line 1690: This equals `eval("$1" + 0)`, I suspect you meant `eval("$1") + 0`

Grinnz: The better way to do this is `tr/_//d`

Leont: Or let `version` take care of it. Or not care at all about the part behind the underscore.

Leont:
This branch really needs a tests that shows what the problem really is. I have a strong feeling it's really the dist that's the problem.

And even if it wasn't the solution sits wrong with me, if anything the code should be simpler not more complex.

lib/ExtUtils/MM_Any.pm
demerphq: > I suspect you meant eval("$1") + 0

Yes. I pushed a fix. I dont see the point in rewrting this to use tr///. Eval is fine IMO.

demerphq:
> This branch really needs a tests that shows what the problem really is.

Agreed. GIven the noises i was hearing from you about this on p5p I didnt want to put much more effort into it beyond proving that we could do better than we are doing. With this patch in place the Slay::Makefile outputs the following.

```
$ perl Makefile.PL 
Generating a Unix-style Makefile
Writing Makefile for Slay::Makefile
Your META.yml has a different license (unknown) than your Makefile.PL (perl_5)
Writing MYMETA.yml and MYMETA.json
```

>if anything the code should be simpler not more complex.

Everthing should be as simple as possible, and no simpler. - Einstein.

It looks like we have set up a complex process that does the wrong thing in a bunch of different ways, it doesn't surprise me that requires to complexity to deal with. The complexity of a solution mirrors the complexity of the problem domain it solves. This problem domain seems complex.

PR #438, opened 2023-03-06
Allow PERL_CORE to be properly overridden with a false value

haarg:
While it was possible to set PERL_CORE to a false value, this would not allow building modules normally if they were in a perl checkout. EUMM would still search for a perl checkout and try to use those paths.

This moves the initialization of PERL_CORE and PERL_SRC to happen earlier, allowing them to be used consistently. And it makes setting PERL_CORE to false skip the PERL_SRC search.


haarg:
I'm not sure if it actually makes sense to try to populate `PERL_SRC` if when `PERL_CORE` is not set. I tried to keep basically the same behavior as previously, but `PERL_CORE` should always be set in the environment when building perl so it seems like it shouldn't be needed. Perhaps it should be replaced with a different check to warn users trying to manually build packages out of a perl source checkout.

Issue #439, opened 2023-03-23
Consider having ExtUtils::MakeMaker maintain all rpath arguments it was given


jkahrman:
ExtUtils::MakeMaker drops any rpath argument that isn't an existing directory. I found that this is too restrictive because:
- the directory could be created after the Makefile is generated
- rpaths can be the variables `$ORIGIN` on Linux or `@rpath`, `@loader_path`, or `@executable_path` on Mac

PR #440, opened 2023-03-23
Proposal to have ExtUtils::MakeMaker maintain all rpath arguments

jkahrman:
ExtUtils::MakeMaker previously dropped any rpath argument that isn't an existing directory. This is too restrictive because:
* the directory could be created after the Makefile is generated
* rpaths can be the variables $ORIGIN on Linux or @rpath, @loader_path, or @executable_path on Mac

Also, do a Makefile escape of any single '$' in the rpath value ($ -> $$). Since the rpath value may also be consumed in a test in Makefile.PL where it doesn't need to be escaped, it's inconvenient to unescape it there.

See #439


lib/ExtUtils/Liblist/Kid.pm
```diff
@@ -89,8 +84,15 @@ sub _unix_os2_ext {
                 elsif ( $Config{'lddlflags'} =~ /-R/ ) {
                     $rtype = '-R';
                 }
+                if ($thislib =~ s{(?<!\$)\$(?!\$)}{\\\$\$}g) {
```
Leont, line 87: This change really ought to be a separate commit

lib/ExtUtils/Liblist/Kid.pm
```diff
@@ -89,8 +84,15 @@ sub _unix_os2_ext {
...
+                    print "Escaping single dollar sign for Makefile\n";
```
Leont, line 88: This should probably only print `if $verbose`

lib/ExtUtils/Liblist/Kid.pm
jkahrman: I had gone back and forth on that since letting $ORIGIN through on the rpaths was the point. I'll break it up.

mohawk2:
I've approved the CI to run.

PR #444, opened 2023-04-15
lib/ExtUtils/MM_Darwin.pm: Unbreak Perl build

sevan:
On modern macOS the compiler defaults to '-Werror,-Wimplicit-function-declaration' This breaks the build on code which calls a function without a prototype (invalid in C99). Turning the check off with '-Wno-error=implicit-function-declaration' breaks the build of  Perl itself on legacy versions of Darwin with compilers which do not recognise the flags. Removing it allows Perl to once again build on Mac OS X as old as 10.4 with GCC 4.0.1.

### DIFF
--- a/lib/ExtUtils/MM_Darwin.pm
+++ b/lib/ExtUtils/MM_Darwin.pm
@@ -46,29 +46,4 @@ sub init_dist {
     $self->SUPER::init_dist(@_);
 }
 
-=head3 cflags
-
-Over-ride Apple's automatic setting of -Werror
-
-=cut
-
-sub cflags {
-    my($self,$libperl)=@_;
-    return $self->{CFLAGS} if $self->{CFLAGS};
-    return '' unless $self->needs_linking();
-
-    my $base = $self->SUPER::cflags($libperl);
-
-    foreach (split /\n/, $base) {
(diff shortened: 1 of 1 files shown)


Leont:
Given that the entire cflags method didn't exist in MM_Darwin until a07d951129587be9c6246e967afc754f4b0cbd68, I'm wondering if we shouldn't remove it entirely. @haarg?

haarg:
Yeah, it looks like the method should just be removed.

sevan:
Ah ok, propagating CFLAGS is no longer a requirement? #8537a20d07008e10179b398f22ffeecf4b1a3d02

haarg:
They will be handled properly by the parent method. The extra work was only needed because they were being modified, which is being removed.

sevan:
The fixes for the affected modules is a trivial one line fix to prototype each defined function. Is there a list of the modules which break?

haarg:
There is not - we should probably look into doing some smoke runs to find out. While a fix may be trivial, getting releases done of all of the modules likely won't be.

haarg (approved):

sevan:
The `dist/ExtUtils-CBuilder` counterpart in Perl has landed. Any chance a release of ExtUtils-MakeMaker can be cut and imported into the perl5 tree so that Perl will once again build out of the box without patching on legacy OS X for the next release?

PR #447, opened 2023-09-18
Fix STDIN input encoding for Windows

hakonhagland:
See https://github.com/cpan-testers/Test-Reporter-Transport-Metabase/issues/3 and https://stackoverflow.com/q/77123411/2173773. I believe this might fix [the issue](https://github.com/cpan-testers/Test-Reporter-Transport-Metabase/issues/3).

### DIFF
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -216,9 +216,18 @@ sub prompt ($;$) {  ## no critic
         print "$def\n";
     }
     else {
+        my $codepage;
+        if ($Is_Win32) {
+            require Win32::Console;
+            require Encode;
+            my $codepage = Win32::GetConsoleCP();  # get the input code page
+        }
         $ans = <STDIN>;
         if( defined $ans ) {
             $ans =~ s{\015?\012$}{};
+            if ($Is_Win32) {
+                $ans = Encode::decode("cp". $codepage, $ans);
+            }
         }
         else { # user hit ctrl-D
             print "\n";


hakonhagland:
Not sure why [this test](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/actions/runs/6220481394/job/16880594194#step:7:64) fails:

```
Use of uninitialized value $codepage in concatenation (.) or string at D:\a\ExtUtils-MakeMaker\ExtUtils-MakeMaker\blib\lib/ExtUtils/MakeMaker.pm line 229.
Unknown encoding 'cp' at D:\a\ExtUtils-MakeMaker\ExtUtils-MakeMaker\blib\lib/ExtUtils/MakeMaker.pm line 229.
# Looks like you planned 11 tests but ran 9.
# Looks like your test exited with 255 just after 9.
t/prompt.t ................ 
```
The test does not fail on my laptop..

jkeenan:
> Not sure why [this test](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/actions/runs/6220481394/job/16880594194#step:7:64) fails:
> 
> ```
> Use of uninitialized value $codepage in concatenation (.) or string at D:\a\ExtUtils-MakeMaker\ExtUtils-MakeMaker\blib\lib/ExtUtils/MakeMaker.pm line 229.
> Unknown encoding 'cp' at D:\a\ExtUtils-MakeMaker\ExtUtils-MakeMaker\blib\lib/ExtUtils/MakeMaker.pm line 229.
> # Looks like you planned 11 tests but ran 9.
> # Looks like your test exited with 255 just after 9.
> t/prompt.t ................ 
> ```
> 
> The test does not fail on my laptop..

Perhaps in your p.r. at line 223 the return value was an empty string.  That would be a defined value, but then at 229 the first argument to `decode` would only be `cp`.

hakonhagland:
> Perhaps in your p.r. at line 223 the return value was an empty string.

@jkeenan Tried to [add a commit](https://github.com/hakonhagland/ExtUtils-MakeMaker/commit/720fd7a8822b58997e266f10494189236abc23d5) to address this, but it does not show up here for some reason..

Leont:
I'm not sure if we can use Encode here like that; MakeMaker is part of perl's bootstrap which means it's used before Encode is built. ExtUtils::MakeMaker::Locale is some prior art in this area.

hakonhagland:
> I'm not sure if we can use Encode here like that

Then as an alternative, would it be possible to apply this patch to e.g. [IO::Prompt::Tiny](https://metacpan.org/pod/IO::Prompt::Tiny) instead and patch consumers like [CPAN::Shell](https://github.com/andk/cpanpm/blob/7931ef4ed69ec68d1dd4de15870d652d621b3bd6/lib/CPAN/Shell.pm#L1621) to use `IO::Prompt::Tiny::prompt` as a replacement for `ExtUtils::MakeMaker::prompt` ?

Leont:
> Then as an alternative, would it be possible to apply this patch to e.g. [IO::Prompt::Tiny](https://metacpan.org/pod/IO::Prompt::Tiny) instead and patch consumers like [CPAN::Shell](https://github.com/andk/cpanpm/blob/7931ef4ed69ec68d1dd4de15870d652d621b3bd6/lib/CPAN/Shell.pm#L1621) to use IO::Prompt::Tiny::prompt as a replacement for ExtUtils::MakeMaker::prompt ?

No. CPAN::Shell is core so can't use any module outside of core.

And I didn't you can't use it at all, I'm just saying it should handle the absence of Encode gracefully.

hakonhagland:
> ExtUtils::MakeMaker::Locale is some prior art in this area.

Yes, nice one! For reference here is a link to the PR from 2014: https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/125

hakonhagland:
> MakeMaker is part of perl's bootstrap which means it's used before Encode is built.

@Leont But why then can `ExtUtils::MakeMaker::Local` use `Encode` itself? See line 15:

https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/blob/694b7c78ab4d505d86a805db9dc0d441f08aabc2/lib/ExtUtils/MakeMaker/Locale.pm#L15

Leont:
> @Leont But why then can `ExtUtils::MakeMaker::Locale` use `Encode` itself? See line 15:

Because [this](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/blob/master/lib/ExtUtils/MakeMaker.pm#L14)

hakonhagland:
> I'm not sure if we can use Encode here like that

Trying with ExtUtils::MakeMaker::Locale instead, see latest commit. Also added a test case.

mohawk2:
I've rebased this against `master` and force-pushed so it gets run against current CI. I didn't touch the commits, but I'd suggest they should be squashed. If the CI likes it, this looks good to me and I'd say it should be merged.

PR #449, opened 2023-10-09
maybe_command(): Ensure that command is defined and non-empty

jkeenan:
As demonstrated in
https://github.com/cpan-testers/CPAN-Reporter/issues/59#issuecomment-1753729292, it is possible to exercise subroutine maybe_command() without a defined argument.  We should preclude this from happening.


haarg:
This doesn't look right to me. The current `maybe_command` method is signaling failure by returning false, but this is making it sometimes throw an exception.

jkeenan:
> This doesn't look right to me. The current `maybe_command` method is signaling failure by returning false, but this is making it sometimes throw an exception.

Correct.  I have revised the subroutine call and the tests.  Please re-review.

lib/ExtUtils/MM_Unix.pm
```diff
@@ -2896,6 +2896,7 @@ Returns true, if the argument is likely to be a command.
 
 sub maybe_command {
     my($self,$file) = @_;
+    $file = '' if (!defined $file or !length $file);
```
ilmari, line 2899: This could be simply `$file = '' unless defined $file;`.

The only value that's `!length` is `''`, so in that case the assignment is redundant.

ilmari: Or even just `return unless defined $file`, since zero-length-name files can't exist.

ilmari, line 2899: This could be more simply

```perl
return unless defined $file and length $file;
```

jkeenan: Yes, that's nice.  Re-pushed.

PR #450, opened 2024-02-04
parse_version: handle "package version" syntax using the "hide from CPAN/PAUSE" hack

eserte:
parse_version cannot parse a "package $package $version" if the line is broken after the "package" keyword, which is usually done for the "hide from CPAN" (or "hide from PAUSE") hack.

Sample affected module:
https://metacpan.org/release/PEVANS/IO-Async-0.803/source/lib/IO/Async/Internals/Connector.pm#L6 (The previous version found in the 0.802 distribution used the traditional package keyword, and there parse_version worked)

@andk @leonerd  FYI

### DIFF
--- a/lib/ExtUtils/MM_Unix.pm
+++ b/lib/ExtUtils/MM_Unix.pm
@@ -3005,16 +3005,32 @@ sub parse_version {
     my($self,$parsefile) = @_;
     my $result;
 
+    my $package_version_rx = qr{ \w[\w\:\']* \s+ (v?[0-9._]+) \s* (;|\{) }x;
+
     local $/ = "\n";
     local $_;
     open(my $fh, '<', $parsefile) or die "Could not open '$parsefile': $!";
     my $inpod = 0;
+    my $maybe_hide_from_cpan = 0;
     while (<$fh>) {
         $inpod = /^=(?!cut)/ ? 1 : /^=cut/ ? 0 : $inpod;
         next if $inpod || /^\s*#/;
         chop;
         next if /^\s*(if|unless|elsif)/;
-        if ( m{^ \s* package \s+ \w[\w\:\']* \s+ (v?[0-9._]+) \s* (;|\{)  }x ) {
+        if ($maybe_hide_from_cpan) {
+            $maybe_hide_from_cpan = 0;
+            if ( m{^ \s* $package_version_rx  }x ) {
+                no warnings;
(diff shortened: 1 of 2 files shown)


eserte:
Some additional context: if CPAN.pm is configured with `allow_installing_module_downgrades=no`, then the versions of the already installed module and the module about to be installed are compared, and the installation is refused if the new one has a lower version, or, as in this case, `undef`.

haarg:
I don't think we should change this. The entire point of using a comment like that is prevent the version from being statically parsed.

This seems like a bug in the IO::Async::Internals::Connector module.

eserte:
The
```
package # hide from CPAN
    Foo::Bar;
```
(the comment could also be written more correctly "hide from PAUSE") hack is widely used to prevent PAUSE to index the module, thus making it a "private" module, so the author is free to do refactoring and remove this module without notice.

The task of parsing a version is orthogonal to this, and should still be possible even when using the new package-version syntax.

Grinnz:
There is a difference here between finding the package to be indexed (which the newline is meant to prevent) and parsing the version. The question is whether fixing parse_version to recognize this would affect what PAUSE chooses to index, as this was not intentionally designed this way.

haarg:
If the module is internal, what is the externally parseable version number for?

karenetheridge:
> If the module is internal, what is the externally parseable version number for?

This is what I would like to know too. What was the usecase where parse_version needed to know the version of IO::Async::Internals::Connector?

eserte:
> This is what I would like to know too. What was the usecase where parse_version needed to know the version of IO::Async::Internals::Connector?

See https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/450#issuecomment-1925893861 . The version comparison is done using `parse_version`.

karenetheridge:
It seems wrong for CPAN.pm to care about the versions of unindexed modules. It really should only be concerned with the main module in the distribution.

mohawk2:
I think it's important that the IO::Async submodule here explicitly says it's supposed to be hidden from CPAN (yes, really it's PAUSE), yet the PR wants to allow CPAN (the module) to see the thing that's trying to be hidden. While one could split a hair that the wording is slightly misleading, I believe @karenetheridge's point is that CPAN.pm is expressly not supposed to see this (as it currently can't) and should stay unable to, which I agree with.

Issue #453, opened 2024-06-10
Link to `perlnewmod` and friends


nataraj-hates-MS-for-stealing-github:
I sometimes create new perl modules for CPAN, but not often enough to remember how did I do it last time.

Usually it takes no time to find `ExtUtils::MakeMaker` but it is not about creating skeleton module distribution that I really need

Finding `perlnewmod` and `module-starter` takes much more time, there are not many mentions in the internet (`ExtUtils::MakeMaker` has many mentions)

I guess it is common mistake, that is made by many potential module authors.

We need for sure link to `perlnewmod` and to `module-starter` and `h2xs`. At least in "SEE ALSO" section.

I would also strongly recommend to add it to "SEE INSTEAD?" section at the beginning of the documentation like it is done in https://metacpan.org/pod/Data::UUID module. To redirect those who is looking for module skeleton creator.

I can do the PR, but I am not able to decide what way it should be.  Can we have "SEE INSTEAD?" section?

nataraj-hates-MS-for-stealing-github:
I added #454 as a suggestion how this can be solved

PR #454, opened 2024-06-10
Adding link to `perlnewmod` to redirect people that needs skeleton module

nataraj-hates-MS-for-stealing-github:
When one is looking for creating new module, he will find ExtUtils::MakeMaker but not `perlnewmod`.

I guess such redirect will make life of many people more easy.

See #453 for more info

### DIFF
--- a/lib/ExtUtils/MakeMaker.pm
+++ b/lib/ExtUtils/MakeMaker.pm
@@ -1370,6 +1370,12 @@ __END__
 
 ExtUtils::MakeMaker - Create a module Makefile
 
+==head1 SEE INSTEAD?
+
+This module is about creating a Makefile for an perl extension module
+from a Makefile.PL. If you are looking for creating skeleton module
+distribution for your new module, please see L<perlnewmod>
+
 =head1 SYNOPSIS
 
   use ExtUtils::MakeMaker;


lib/ExtUtils/MakeMaker.pm
```diff
@@ -1370,6 +1370,12 @@ __END__
 
 ExtUtils::MakeMaker - Create a module Makefile
 
+==head1 SEE INSTEAD?
```
haarg, line 1373: There's an extra `=` here.

haarg (commented):
This seems to overlap with the intent of the "[How To Write A Makefile.PL](https://metacpan.org/pod/ExtUtils::MakeMaker#How-To-Write-A-Makefile.PL)" section. Putting the section above DESCRIPTION does make sense I think, but I'm not sure if `perlnewmod` alone is all that useful.

I could see a section linking to `ExtUtils::MakeMaker::Tutorial`, `perlnewmod`, and `Module::Starter` that is above `DESCRIPTION`. And possibly some additional changes to `ExtUtils::MakeMaker::Tutorial` to better promote tools like `Module::Starter`, `Dist::Zilla`, `h2xs`, etc.

PR #456, opened 2024-06-20
set status as unstable for TRIAL releases

haarg:


karenetheridge (approved):

Leont (approved):

Leont:
Perhaps we should add a `TRIAL=1` argument for this?

bingos:
I think $self->{DISTVNAME} needs to be checked for definedness.

```
diff --git a/lib/ExtUtils/MM_Any.pm b/lib/ExtUtils/MM_Any.pm
index 27940912..b70775d7 100644
--- a/lib/ExtUtils/MM_Any.pm
+++ b/lib/ExtUtils/MM_Any.pm
@@ -1302,7 +1302,7 @@ sub metafile_data {
     $meta_merge ||= {};
 
     my $version = _normalize_version($self->{VERSION});
-    my $unstable = $version =~ /_/ || $self->{DISTVNAME} =~ /-TRIAL\d*$/;
+    my $unstable = $version =~ /_/ || ( defined $self->{DISTVNAME} and $self->{DISTVNAME} =~ /-TRIAL\d*$/ );
     my $release_status = $unstable ? 'unstable' : 'stable';
     my %meta = (
         # required
```
That fixes it for me.

Leont:
> That fixes it for me.

I don't think that is the right fix. `DISTVNAME` will only ever be undefined in that specific test because that constructs an incomplete MakeMaker object instead of using the actual constructor. A test-only issue should be fixed in the test instead.

PR #468, opened 2025-03-25
WriteEmptyMakefile should produce the same target types as WriteMakefile

haarg:
The target types created by WriteEmptyMakefile should match the ones created by WriteMakefile. This would ensure that any munging of the file to add rules could operate the same on each.


mohawk2:
This is a great idea! It would be even more perfect if the "empty" file also had a `clean`/`realclean` for maximum tidiness; a thing I've felt a need for back when PDL had lots of optional-build (with empty makefiles left lying around after `realclean`).

PR #474, opened 2025-07-12
Support 'class' VERSIONs, like 'package'

ehuelsmann:
This is supported by Perl since 5.38.


ehuelsmann:
The consequence of 13d823 is that the first attribute of a `class` declaration needs to be on the same line as the class keyword itself. Is that acceptable?

